PR #13, opened 2015-10-25
Fix big/little endian alignment [UNTESTED!]

toxxin:
Hello guys!
We faced with big/little-endians [issue](https://github.com/openwrt/packages/pull/1860). At least it looks like it is. Could you test and merge this patch please, if this bug confirmed? Sorry, I can't test it properly, bc I have only little-endian arch SoCs.
Thanks!

Signed-off-by: Anton Glukhov anton.a.glukhov@gmail.com

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,6 +27,8 @@ AM_PROG_CC_C_O
 AM_PROG_AR
 AC_CONFIG_MACRO_DIR([m4])
 
+AC_C_BIGENDIAN
+
 # enable the automatically build of shared and static libraries, the unit
 # tests needs the static library
 LT_INIT([shared static])


obgm:
Thanks, I will fix this asap. It seems that the entire endianness checks have been dropped when the build process was restructured. There is also another important piece missing that will set WORDS_BIGENDIAN used in pdu.h. My proposed way to fix this in a cleaner way is to serialize the coap header manually before sending (coap_pdu_parse() already does this when reading a packet from the network).

rasendubi:
Tested the patch on powerpc32. Works fine.

anupam19:
I think it affects Arduino Yun too, in my case I saw malformed packet (coming from Yun) warning at Wireshark.

malinengineer:
@anupam19 You are probably right. Arduino Yun is based on the AR9331 SoC, which seems to be big endian.

anupam19:
@malinengineer true that :+1:
http://www.datasheetspdf.com/PDF/AR9331/743003/24
http://forum.arduino.cc/index.php?topic=241368.15

obgm:
After having tested it again on a mips64-octeon platform, I have pulled in the patch and close this PR.

PR #32, opened 2016-06-09
configure: only add -Wunused-result if supported by the compiler

tpetazzoni:
Old gcc versions such as 4.3.x do not support -Wunused-result, so
instead of unconditionally using this warning, only use it if supported
by the compiler.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -110,12 +110,12 @@ WARNING_CFLAGS="\
 -Wswitch-default \
 -Wswitch-enum \
 -Wunused \
--Wunused-result \
 "
 
 # check whether or not the compiler supports -Wlogical-op (clang does not...)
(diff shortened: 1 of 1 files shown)


tijuca:
Hi Thomas,

I'm fine with this PR, thanks for tuning this part. The above CI error isn't related to this change as it's clearly to see that clang and GCC is making the check successfully and the issue is related to Contiki code.
```
checking whether C compiler accepts -fdiagnostics-color... yes
checking whether C compiler accepts -Wunused-result... yes   <-----
checking whether the linker accepts -Wl,--version-script=./libcoap-1.map... yes
```

obgm:
Thanks @tpetazzoni and @tijuca -- I have pulled this PR into develop.

PR #36, opened 2016-06-24
Add missing LOG_ERR log level

Guimier:
The fallback used when &lt;syslog.h&gt; is not available was missing the error level.


obgm:
Applied, thanks!

PR #41, opened 2016-07-14
Add documentation for flags used in resources creation.

KurdyMalloy:
Add documentation based on a question in the mailing list.

### DIFF
--- a/include/coap/resource.h
+++ b/include/coap/resource.h
@@ -57,8 +57,19 @@ typedef struct coap_attr_t {
   int flags;
 } coap_attr_t;
 
+/** The URI passed to coap_resource_init() is free'd by coap_delete_resource(). */
 #define COAP_RESOURCE_FLAGS_RELEASE_URI 0x1
+
+/**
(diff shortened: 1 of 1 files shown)


obgm:
Thank you. Can you please replace the '-' in RFC-7641 by a white space as this is more common?
By ' 4.5', you probably mean ', Section 4.5'?
I wonder if you could include a hyperlink for more convenient access to that document.

tijuca:
And please take care of paragraph 5.5 from [CONTRIBUTE](https://github.com/obgm/libcoap/blob/develop/CONTRIBUTE#L161) and add some prefix to the commit topic, this helps not only me to go through all commits if needed.

KurdyMalloy:
I tried to squash everything in one commit with the correct topic which resulted in 3de2558, the merge (cc12a86) is a result of my lack of talent with git I suppose ;-) , I assume you can just cherry-pick 3de2558.

obgm:
Fixed in 02c3e67.

PR #42, opened 2016-07-15
Document flags option in coap_resource_init

nikosft:
Possible values for the flags option have been added in the declaration of coap_resource_init. This change will also affect the generated doxygen documentation

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -330,7 +330,7 @@ PREDEFINED_CFLAGS=`echo $ADDITIONAL_CFLAGS | sed -e 's/-D//g'`
 AC_SUBST(PREDEFINED_CFLAGS)
 
 # And finaly combining the CFLAGS together ...
-CFLAGS+=" $ADDITIONAL_CFLAGS"
+CFLAGS="$CFLAGS $ADDITIONAL_CFLAGS"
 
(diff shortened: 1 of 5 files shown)


obgm:
Thank you. Please combine both submissions into one commit. Moreover, I recommend mentioning that COAP_RESOURCE_FLAGS_NOTIFY_NON is the default (i.e., when flag == 0).

nikosft:
I added in the description your recommendation. Nevertheless, I am not sure how to merge commits and I believe I made things worse.

obgm:
Okay, that one resulted in a merge conflict that sneaked into your commit. Usually, you could try rewriting history on your local working copy using (attention, dangerous operations follow, handle with care:) `git reset --hard` and possibly even (more dangerous) git reflog.
I am not sure if repositories that have been pushed can be cleared up that way. I suppose not.
What I can offer as an option is that I pick from your fork the relevant changes in patch format and commit that on your behalf. This might be easier but may mess up your fork even more.

obgm:
Actually, I would prefer to have @n instead of <br> to enforce the line breaks to be more or less independent of HTML.

nikosft:
I think it is easier to create a new fork and make changes there. I am closing this pull request and I will create a new one.

PR #43, opened 2016-07-18
flags option in coap_resource_init documented

nikosft:
Possible values of the flags option of the coap_resource_init have been documented. This change also affects the documentation generated by doxygen

### DIFF
--- a/include/coap/resource.h
+++ b/include/coap/resource.h
@@ -102,8 +102,12 @@ typedef struct coap_resource_t {
  *
  * @param uri    The URI path of the new resource.
  * @param len    The length of @p uri.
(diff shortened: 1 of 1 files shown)


tijuca:
Thanks for taking care.
Some minor point, I'd like to see some prefix in the commit header message like here something like `doxygen:` or `resource.h:` Please see file CONTRIBUTE paragraph 5.5.
if you have to go through a lot of commits this helps a lot to ignore not relevante commits.

tijuca:
@obgm 
What's your thinking on this old PR. It's a bit sad we are not keep such more simple changes by amending them locally.

obgm:
Done.

PR #47, opened 2016-09-17
Changes from Iotivity for draft-ietf-core-coap-tcp-tls and cross-platform support

dthaler:
These changes were already in the Iotivity code base, and we're trying to unfork libcoap and have iotivity just use the regular libcoap library.   These changes include support for draft-ietf-core-coap-tcp-tls and some cross-platform fixes (e.g., to also run on Windows).

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -69,7 +69,7 @@ coap_tick_t obs_wait = 0;               /* timeout for current subscription */
 #define UNUSED_PARAM
 #endif /* GCC */
 
-static inline void
+COAP_STATIC_INLINE void
 set_timeout(coap_tick_t *timer, const unsigned int seconds) {
   coap_ticks(timer);
   *timer += seconds * COAP_TICKS_PER_SECOND;
@@ -305,7 +305,7 @@ resolve_address(const str *server, struct sockaddr *dst) {
    ((Pdu)->hdr->code == COAP_RESPONSE_CODE(201) ||                \
     (Pdu)->hdr->code == COAP_RESPONSE_CODE(204)))
 
-static inline int
+COAP_STATIC_INLINE int
 check_token(coap_pdu_t *received) {
   return received->hdr->token_length == the_token.length &&
     memcmp(received->hdr->token, the_token.s, the_token.length) == 0;
@@ -808,7 +808,7 @@ cmdline_proxy(char *arg) {
   return 1;
 }
 
-static inline void
+COAP_STATIC_INLINE void
 cmdline_token(char *arg) {
   strncpy((char *)the_token.s, arg, min(sizeof(_token_data), strlen(arg)));
   the_token.length = strlen(arg);
--- a/include/coap/pdu.h
+++ b/include/coap/pdu.h
@@ -248,15 +248,17 @@ typedef struct {
   unsigned char header_data[6];
   unsigned char token[]; /* the actual token, if any */
 } coap_hdr_tcp_32bit_t;
+#endif /* WITH_TCP */
 
 typedef union {
   coap_hdr_udp_t udp;
+#ifdef WITH_TCP
   coap_hdr_tcp_t tcp;  
   coap_hdr_tcp_8bit_t tcp_8bit;
   coap_hdr_tcp_16bit_t tcp_16bit;
   coap_hdr_tcp_32bit_t tcp_32bit;
-} coap_hdr_transport_t;
 #endif /* WITH_TCP */
+} coap_hdr_transport_t;
 
 // Typedef for backwards compatibility.
 typedef coap_hdr_udp_t coap_hdr_t;
@@ -299,12 +301,10 @@ typedef struct {
                                *   This may or may not equal (coap_hdr_t*)(pdu+1)
                                *   depending on the memory management
                                *   implementation. */
-#ifdef WITH_TCP
     coap_hdr_transport_t *transport_hdr; /**< Address of the first byte of the CoAP message.
                                            *   This may or may not equal (coap_hdr_t*)(pdu+1)
                                            *   depending on the memory management
                                            *   implementation. */
-#endif
   };
   unsigned short max_delta; /**< highest option number */
   unsigned short length;    /**< PDU length (including header, options, data) */
@@ -367,12 +367,40 @@ coap_pdu_init(unsigned char type,
               size_t size);
 
 /**
+ * Creates a new CoAP PDU of given @p size (must be large enough to hold the
+ * basic CoAP message header (coap_hdr_t). The function returns a pointer to
+ * the node coap_pdu_t object on success, or @c NULL on error. The storage
+ * allocated for the result must be released with coap_delete_pdu().
+ *
+ * @param type The type of the PDU (one of COAP_MESSAGE_CON,
+ *             COAP_MESSAGE_NON, COAP_MESSAGE_ACK, COAP_MESSAGE_RST).
+ * @param code The message code.
+ * @param id   The message id to set or COAP_INVALID_TID if unknown.
+ * @param size The number of bytes to allocate for the actual message.
+ * @param transport The transport type.
+ *
+ * @return A pointer to the new PDU object or @c NULL on error.
+ */
+coap_pdu_t *
+coap_pdu_init2(unsigned char type, unsigned char code, unsigned short id,
+               size_t size, coap_transport_t transport);
+
+/**
  * Clears any contents from @p pdu and resets @c version field, @c
  * length and @c data pointers. @c max_size is set to @p size, any
  * other field is set to @c 0. Note that @p pdu must be a valid
  * pointer to a coap_pdu_t object created e.g. by coap_pdu_init().
  */
 void coap_pdu_clear(coap_pdu_t *pdu, size_t size);
+
+/**
+ * Clears any contents from @p pdu and resets @c version field, @c
+ * length and @c data pointers. @c max_size is set to @p size, any
+ * other field is set to @c 0. Note that @p pdu must be a valid
+ * pointer to a coap_pdu_t object created e.g. by coap_pdu_init().
+ */
+void coap_pdu_clear2(coap_pdu_t *pdu, size_t size, coap_transport_t transport,
+                     unsigned int length);
 
 /**
  * Creates a new CoAP PDU.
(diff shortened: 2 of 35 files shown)


obgm:
Thanks, this contribution is highly appreciated. Unfortunately, Travis reports this PR to break the Linux port. It would be great if you could take a look into that issue.

dthaler:
Just pushed a fix for the bug due to a bad merge

obgm:
My apologies -- I have pushed an old patch to src/block.c which now breaks this PR.

src/net.c
```diff
@@ -554,7 +554,7 @@ coap_send_ack(coap_context_t *context,
   return result;
 }
 
-#if defined(WITH_POSIX) || defined(WITH_CONTIKI)
+#if !defined(WITH_LWIP)
```
tijuca: Please don't use negative logic here and let that check as it is, it's hard to figure out some not well knowledge people what !WITH_LWIP means!

rzr:
Is this task tracked on iotivity tracker ?

dthaler:
Yes https://jira.iotivity.org/browse/IOT-1072 tracks the iotivity task.  IoTivity already uses this branch temporarily on Windows and other OS's will start snapping to it after merge.   I will rebase to fix the conflict.

dthaler:
I believe all feedback so far has been addressed.  Would like to see this PR merged so iotivity can start depending on the upstream branch.  Let me know if there are any more blocking issues.

tijuca:
Hello Dave,

first thanks for contributing! Some comments below.

Beside the build issues I'd like to see better commit messages, please see and note https://github.com/obgm/libcoap/blob/develop/CONTRIBUTE#L161
Please describe why the commit is made in the way you have done and make explanation what the commit is about (for example c1bda6e0a4136bc3a3f981ffb4d38a984f2a4f3c 749514f4200cc0675dc134562b42a8ca06b831fe fd5c9ec1f584c06fc1608df61e32fe4694c1b0df). It's a bit tedious to go through every source change to understand what's changed and why. That's obviously not needed for typo fixes, but a good commit message helps to see what was changed.

Please don't mix extensions of specific changes to the header files or small adoptions within the header files with improving of the API documentation in other parts, split them of into more easier to understand changes.
For example c11591d5237dfd8880971b319eabab8d81db7e9b is much to big. But other commits will need also be split of into smaller peaces.

Also try to decrease merge commit from upstream much as possible, normally there should be even one as you should use `git rebase` instead of `git merge` to pull upstream changes. I'm currently not able to make a review due the huge changes that I not really understand what and why the source is changed.

Please don't be disappointed, such big changes need time to get integrated. Maybe you can split the PR into more than one PR if it turns out there are some thematic changes. If I read your first post this PR is including two things, tcp-tls and cross-platform support for Windows. For the latter it would be good to have build support than.

Regards
Carsten

obgm:
Hi Dave, I had a quick look at the code and would like to process it quickly. Do you see a chance to address Carsten's comments by splitting this PR into multiple commits?

obgm:
Building with -DWITH_TCP=1 (after fixing the build errors) yields even more warnings.

examples/client.c
```diff
@@ -12,18 +12,37 @@
...
+#endif
+#ifdef HAVE_ARPA_INET_H
+# include <arpa/inet.h>
+#endif
+#ifdef HAVE_NETDB_H
+# include <netdb.h>
+#endif
+#ifndef HAVE_STRCASECMP
+  /* The POSIX strcasecmp() is not supported, so use the ISO C++ conformant _stricmp instead. */
+# define strcasecmp _stricmp
```
obgm: I wonder where _stricmp() might be defined. In my (draft version) of ISO 14882:2011, there is no such thing. Please enlighten me.

examples/client.c
```diff
@@ -694,7 +715,7 @@ cmdline_uri(char *arg, int create_uri_opts) {
   if (proxy.length) {   /* create Proxy-Uri from argument */
     size_t len = strlen(arg);
     while (len > 270) {
-      coap_insert(&optlist,
+      coap_append(&optlist,
```
obgm: See above for notes on these changes.

src/block.c
```diff
@@ -52,6 +52,8 @@ coap_get_block(coap_pdu_t *pdu, unsigned short type, coap_block_t *block) {
   memset(block, 0, sizeof(coap_block_t));
 
   if (pdu && (option = coap_check_option(pdu, type, &opt_iter)) != NULL) {
+    unsigned int num;
```
obgm: This has been present before. Did you rebase to the HEAD?

include/coap/pdu.h
```diff
@@ -53,11 +58,14 @@
 #define COAP_REQUEST_DELETE    4
 
 /* CoAP option types (be sure to update check_critical when adding options */
+/* See http://www.iana.org/assignments/core-parameters/core-parameters.xhtml#option-numbers */
 
 #define COAP_OPTION_IF_MATCH        1 /* C, opaque, 0-8 B, (none) */
 #define COAP_OPTION_URI_HOST        3 /* C, String, 1-255 B, destination address */
 #define COAP_OPTION_ETAG            4 /* E, opaque, 1-8 B, (none) */
 #define COAP_OPTION_IF_NONE_MATCH   5 /* empty, 0 B, (none) */
+#define COAP_OPTION_OBSERVE         6 /* E, empty/uint, 0 B/0-3 B, (none) */
```
obgm: Please, keep the previous order (clustered by specification document -- the option numbers are irrelevant for ordering).

include/coap/pdu.h
```diff
@@ -171,28 +180,94 @@ typedef int coap_tid_t;
...
+    COAP_TCP,
+    COAP_TCP_8BIT,
+    COAP_TCP_16BIT,
+    COAP_TCP_32BIT
+#endif
+} coap_transport_t;
+
 #ifdef WORDS_BIGENDIAN
 typedef struct {
-  unsigned int version:2;      /* protocol version */
```
obgm: Please, keep the type unsigned int as this (besides int) is the only type allowed for bit fields (hence the corresponding compiler warnings).

include/coap/pdu.h
```diff
@@ -226,10 +301,16 @@ typedef struct {
...
-  coap_hdr_t *hdr;          /**< Address of the first byte of the CoAP message.
-                             *   This may or may not equal (coap_hdr_t*)(pdu+1)
-                             *   depending on the memory management
-                             *   implementation. */
+  union {
+    coap_hdr_t *hdr;          /**< Address of the first byte of the CoAP message.
+                               *   This may or may not equal (coap_hdr_t*)(pdu+1)
+                               *   depending on the memory management
+                               *   implementation. */
+    coap_hdr_transport_t *transport_hdr; /**< Address of the first byte of the CoAP message.
```
obgm: Here, the UDP header is included twice.

include/coap/pdu.h
```diff
@@ -308,6 +417,15 @@ void coap_pdu_clear(coap_pdu_t *pdu, size_t size);
  */
 coap_pdu_t *coap_new_pdu(void);
 
+/**
+ * Creates a new CoAP PDU. The object is created on the heap and must be released
+ * using coap_delete_pdu();
+ *
+ * @deprecated This function allocates the maximum storage for each
```
obgm: As the original function has been clearly marked as deprecated, I do not think that a new function should be introduced that mimics the deprecated behavior.

include/coap/pdu.h
```diff
@@ -308,6 +417,15 @@ void coap_pdu_clear(coap_pdu_t *pdu, size_t size);
...
+ * PDU. Use coap_pdu_init2() instead.
+ */
+coap_pdu_t *coap_new_pdu2(coap_transport_t transport, unsigned int size);
```
obgm: Regarding the general transport-specific API: I think that it is not a good idea to specify the transport for each and every operation on a PDU. When only UDP existed as transport for CoAP, you could safely assume that the CoAP message under construction will have a certain serialization on the wire. Now, when introducing multiple transports, the serialization should happen immediately when writing the message onto the network. The API for setting the token, adding options or data, etc., should not have to care which serialization will be used,

include/coap/pdu.h
```diff
@@ -226,10 +301,16 @@ typedef struct {
...
+  union {
+    coap_hdr_t *hdr;          /**< Address of the first byte of the CoAP message.
+                               *   This may or may not equal (coap_hdr_t*)(pdu+1)
+                               *   depending on the memory management
+                               *   implementation. */
+    coap_hdr_transport_t *transport_hdr; /**< Address of the first byte of the CoAP message.
+                                           *   This may or may not equal (coap_hdr_t*)(pdu+1)
+                                           *   depending on the memory management
+                                           *   implementation. */
+  };
```
obgm: Do not use anonymous unions (hence the compiler warnings).

dthaler:
FYI, The changes in this PR have been done in IoTivity's fork of this repo across the course of a couple years by a bunch of people.  I am merely pushing them here, I am not authoritative for all the original changes, just trying to get rid of a forked copy and make contributions in the future to a single place rather than IoTivity folks continuing to make changes to their own fork.   I will do what I can, but I may not be able to answer all questions myself.   Pushing back just means that tons of devices (like all the devices Samsung ships) will continue to use a separate fork not the master GitHub version.

rzr:
Hi as Samsung OSG developer I believe it's better for iotivity to be the most aligned possible with upstream, thanks @dthaler for your efforts.

tijuca:
Dave, thanks for a little bit explanation about the details of this PR. 
I agree and encourage the best for libcoap **and** the other projects is obviously to land pull requests early in libcoap and the uses code is not differing much between projects. It's like in the kernel, commit early and do also a pull request early.

I think Olaf is of course willing to pull your changes in this PR. But we need to puzzle the various things out into parts that are logical things. That for one simple reason, not only we need to understand what the changes are about. So how to get further here?

I'd suggest to do the easy parts first, for example changes that are made by you or direct working people.

Olaf has released the recent state of libcoap as version 4.1.2 so we can start now mostly lazy with new changes.

AnderssonPeter:
Any news on this?
I would love to have Windows support!

obgm:
It certainly is high on my priority list but sorting stuff out will take some time.
Maybe it is feasible to cherry-pick some of the Windows-specific compatibility code. That part seems pretty straight-forward to me.

obgm:
@dthaler I am having trouble understanding what these changes are about. For example, commit 5f2d2386 introduces a function `coap_add_token_to_empty_message()`. Neither the commit log nor the doxygen comment for the function prototype in commit bd60461d explain what this function is about and, more important, why it is needed in addition to `coap_add_token()`.

PR #49, opened 2016-09-29
Change some wrong response code value to match the good phrase

jpcomtois:
There seems to have a mismatch with some actual response code value, as :
65 is the answer of COAP_RESPONSE_CODE(201)...
66 is the answer of COAP_RESPONSE_CODE(202)...
etc..
In this way, COAP_RESPONSE_CODE(65) should be COAP_RESPONSE_CODE(201) to fit the corresponding phrase.

### DIFF
--- a/src/pdu.c
+++ b/src/pdu.c
@@ -279,11 +279,11 @@ typedef struct {
 /* if you change anything here, make sure, that the longest string does not 
  * exceed COAP_ERROR_PHRASE_LENGTH. */
 error_desc_t coap_error[] = {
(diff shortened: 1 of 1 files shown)


tijuca:
@obgm 

What about this change? I'm not that deep in the RFCs but in my eyes this fix would make sense. I'd like see if the commit could be prefixed like something _pdu.c: correct wrong response code for 20x_ and adding some more explaining body in case the change is correct.

obgm:
Thanks. I have included this with a slightly edited commit message as 718b2df.

PR #52, opened 2016-11-06
Allow client example to not generate Uri-Host and Uri-Port options

kb2ma:
Added a new command line option, -U.

I work with constrained implementations of CoAP, [gcoap](https://github.com/RIOT-OS/RIOT/pull/5598) and [nanocoap](https://github.com/kaspar030/sock/tree/master/nanocoap). I appreciate having the examples in libcoap to test against. However, these constrained implementations have no need for the Uri-Host or Uri-Port options -- there always is a single host and port being served. Since these options are 'critical', the implementation can't ignore them, so I would like the ability to ensure they will not be sent.


obgm:
Thanks!

PR #55, opened 2016-11-29
debug: Fix off-by-1 error in coap_show_pdu block size output

gebart:
The error made the debug printout show a block size which is twice as large as the actual block size set by the CoAP option value.


obgm:
Thanks, this is now in commit ed6ca0e.

PR #56, opened 2016-11-29
examples/client: Fix Block1 handling for growing block sizes

gebart:
Only increment block size if it can be done without causing gaps or overlaps in the byte sequence.

The patch adds a check for alignment to ensure that no gaps appear when a server responds with a larger block size than the original request was using.

Ignore the block number sent by the server since that may or may not be valid. Always recompute based on how many bytes have been sent so far.

Rationale:

Erbium (Contiki CoAP) responds to Block1 request with a Block1 option containing the server's maximum supported Block1 size (Contiki configuration option `REST_MAX_CHUNK_SIZE`). However, the block number in the Block1 option is the same number as the block number of the original request.

I'm not sure Contiki is correct in its use of the block number in its Block1 responses, but it feels like it will be more robust to do like this and recompute the block number whenever the block size changes.

### DIFF
--- a/examples/client.c
+++ b/examples/client.c
@@ -426,11 +426,24 @@ message_handler(struct coap_context_t *ctx,
       block_opt = coap_check_option(received, COAP_OPTION_BLOCK1, &opt_iter);
 
       if (block_opt) { /* handle Block1 */
-        block.szx = COAP_OPT_BLOCK_SZX(block_opt);
-        block.num = coap_opt_block_num(block_opt);
-
-        debug("found Block1, block size is %u, block nr. %u\n",
-        block.szx, block.num);
+        unsigned int szx = COAP_OPT_BLOCK_SZX(block_opt);
+        unsigned int num = coap_opt_block_num(block_opt);
+        debug("found Block1 option, block size is %u, block nr. %u\n", szx, num);
+        if (szx != block.szx) {
+          unsigned int bytes_sent = ((block.num + 1) << (block.szx + 4));
+          if (bytes_sent % (1 << (szx + 4)) == 0) {
+            /* Recompute the block number of the previous packet given the new block size */
+            block.num = (bytes_sent >> (szx + 4)) - 1;
+            block.szx = szx;
+            debug("new Block1 size is %u, block number %u completed\n", (1 << (block.szx + 4)), block.num);
+          } else {
+            debug("ignoring request to increase Block1 size, "
+            "next block is not aligned on requested block size boundary. "
+            "(%u x %u mod %u = %u != 0)\n",
+                  block.num + 1, (1 << (block.szx + 4)), (1 << (szx + 4)),
+                  bytes_sent % (1 << (szx + 4)));
+          }
+        }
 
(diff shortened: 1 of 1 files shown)


examples/client.c
```diff
@@ -426,11 +426,24 @@ message_handler(struct coap_context_t *ctx,
...
-        block.szx = COAP_OPT_BLOCK_SZX(block_opt);
-        block.num = coap_opt_block_num(block_opt);
-
-        debug("found Block1, block size is %u, block nr. %u\n",
-        block.szx, block.num);
+        unsigned int szx = COAP_OPT_BLOCK_SZX(block_opt);
+        unsigned int num = coap_opt_block_num(block_opt);
+        debug("found Block1 option, block size is %u, block nr. %u\n", szx, num);
+        if (szx != block.szx) {
+          unsigned int bytes_sent = (block.num + 1) * (1 << (block.szx + 4));
```
obgm: The multiplication seems a bit redundant. How about this:

```c
unsigned int bytes_sent = (block.num + 1) << (block.szx + 4);
```

examples/client.c
```diff
@@ -426,11 +426,24 @@ message_handler(struct coap_context_t *ctx,
...
+          if (bytes_sent % (1 << (szx + 4)) == 0) {
```
obgm: The modulus could be replaced by bit-wise and because we are dealing with powers of 2:

```c
if (bytes_sent & ((1 << szx + 4) - 1) {
```

obgm:
Thanks @gebart for this PR. I think you got this right.  I have a few small remarks included with my review. Please check if that makes sense to you.

examples/client.c
gebart: I agree, it looks a bit redundant. I'll change it.

examples/client.c
gebart: Replacing the code with a bitwise and would be a premature optimization in my opinion. I applied your suggested change and compared the disassembly at `-O2` optimization level, the output is identical. The point is that the compiler already knows about this optimization and we don't have to apply it manually, the change will only make the code slightly more difficult to read for users less used to working with bitwise operations

GCC version: `gcc (Gentoo 5.4.0 p1.0, pie-0.6.5) 5.4.0`
Command line (default configure settings): `gcc -DHAVE_CONFIG_H -I. -I..    -isystem../include/coap -I../include/coap -pedantic -Wall -Wextra -Wformat-security -Winline -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wshadow -Wstrict-prototypes -Wswitch-default -Wswitch-enum -Wunused  -Wlogical-op -Wunused-result -std=c99 -g -O2 -fdiagnostics-color -D_GNU_SOURCE -DWITH_POSIX -MT client.o -MD -MP -MF .deps/client.Tpo -c -o client.o client.c`)

gebart:
Removed the redundant multiplication, amended original commit (minor edit).

examples/client.c
obgm: Fair enough, you may leave that as is if you want to.

PR #65, opened 2017-03-08
Fixing Option Delta and Option Length encoding.

PawelWMS:
Option Delta and Option Length were incorrectly encoding
the value 269 and ended up being encoded the same way as 13.

For details please see RFC-7252 "Option Format"
( https://tools.ietf.org/html/rfc7252#page-17 ).


obgm:
Thank you very much!

PR #67, opened 2017-03-11
Fixing coap_remove_failed_observers() loop.

PawelWMS:
The loop was always exiting after analysing the first observer,
so in case there were more and the first one on the list of
observers didn't match the one we wanted to remove, the
observer never got removed.


obgm:
You are right, the break statement was misplaced.

PR #70, opened 2017-03-20
Fixing /W4 warnings for Windows builds.

PawelWMS:
Removing all of the /W4 level warnings from the code.
No functional changes.

A side effect of fixing the warnings is also making sure all
length/size casts don't end up with data loss.

### DIFF
--- a/src/coap_io.c
+++ b/src/coap_io.c
@@ -258,6 +258,11 @@ coap_network_send(struct coap_context_t *context UNUSED_PARAM,
 #ifndef WITH_CONTIKI
   /* a buffer large enough to hold all protocol address types */
   char buf[CMSG_LEN(sizeof(struct sockaddr_storage))];
+  (void)datalen;
+  (void)data;
+  (void)context;
+
+  mhdr.Control.buf = buf;
   struct msghdr mhdr;
   struct iovec iov[1];
 
@@ -387,6 +392,7 @@ coap_free_packet(coap_packet_t *packet) {
 
 static inline size_t
 coap_get_max_packetlength(const coap_packet_t *packet UNUSED_PARAM) {
+  (void)packet;
   return COAP_MAX_PDU_SIZE;
 }
 
@@ -610,10 +616,12 @@ coap_network_read(coap_endpoint_t *ep, coap_packet_t **packet) {
   (*packet)->interface = ep;
 
   return len;
+#if defined(WITH_POSIX) || defined(WITH_CONTIKI)
  error:
   coap_free_packet(*packet);
   *packet = NULL;
   return -1;
+#endif
 }
 
 #undef SIN6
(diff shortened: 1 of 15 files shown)


PawelWMS:
@obgm: Please take a look at this change - it would be

src/coap_io.c
```diff
@@ -258,6 +258,11 @@ coap_network_send(struct coap_context_t *context UNUSED_PARAM,
 #ifndef WITH_CONTIKI
   /* a buffer large enough to hold all protocol address types */
   char buf[CMSG_LEN(sizeof(struct sockaddr_storage))];
+  (void)datalen;
```
PawelWMS: You're right - sorry about that. It was a result of a merge issue I missed. I moved this line and the next to the "else" part of "#ifndef WITH_CONTIKI".

src/coap_time.c
```diff
@@ -47,7 +48,7 @@ coap_clock_init(void) {
 
 void
 coap_ticks(coap_tick_t *t) {
-  unsigned long tmp;
+  coap_tick_t tmp;
```
PawelWMS: Another merge issue - this change is not needed here, sorry for that.

Just for the record *tmp* would still be an *unsigned long*, because if *WITH_POSIX* is defined you end up with *typedef unsigned long coap_tick_t* inside coap_time.h.

src/mem.c
```diff
@@ -32,11 +32,13 @@ coap_memory_init(void) {
 
 void *
 coap_malloc_type(coap_memory_tag_t type UNUSED_PARAM, size_t size) {
+  (void)type;
```
PawelWMS: If you'd like, I could define something like a *COAP_UNUSED*() macro, to make the intentions more clear.

src/net.c
```diff
@@ -494,6 +494,10 @@ coap_option_check_critical(coap_context_t *ctx,
 void
 coap_transaction_id(const coap_address_t *peer, const coap_pdu_t *pdu, 
 		    coap_tid_t *id) {
+#if !defined(WITH_POSIX) && !defined(WITH_LWIP) && !defined(WITH_CONTIKI)
+  (void)peer;
```
PawelWMS: In that case we could make the whole *coap_transaction_id*() unavailable if neither of these *WITH_** macros is defined. Otherwise *peer* is not used anyway and the result doesn't change.

src/net.c
```diff
@@ -1557,6 +1569,8 @@ handle_locally(coap_context_t *context __attribute__ ((unused)),
 handle_locally(coap_context_t *context, coap_queue_t *node) {
 #endif /* GCC */
   /* this function can be used to check if node->pdu is really for us */
+  (void)context;
```
PawelWMS: I can remove the conditional, but is this function really needed, since it always returns 1, regardless of the platform and arguments?

src/net.c
obgm: Yes, you are right. Seems to be the better way.

src/net.c
obgm: Indeed. It seemed a good idea at that time (2011!) but was never used for anything, so we can just get rid of it.

obgm:
Thanks for the quick changes. I think, we will have to cleanup the #ifdefs from net.h but for now, this is good enough to pull in and close this PR.

PR #76, opened 2017-04-04
net.c: correct brackets in if-statement

PawelWMS:
Making sure that "if" statements have opening brackets
on the same line to fit with the rest of the code.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1255,8 +1255,7 @@ coap_wellknown_response(coap_context_t *context, coap_pdu_t *request) {
   }
   
   unsigned int new_resp_length = resp->length + COAP_PRINT_OUTPUT_LENGTH(result);
-  if (new_resp_length > USHRT_MAX)
-  {
+  if (new_resp_length > USHRT_MAX) {
       debug("coap_print_wellknown failed - print result too large\n");
       goto error;
   }
--- a/src/resource.c
+++ b/src/resource.c
@@ -282,8 +282,7 @@ coap_print_wellknown(coap_context_t *context, unsigned char *buf, size_t *buflen
   *buflen = written;
   output_length = p - buf;
 
-  if (output_length > COAP_PRINT_STATUS_MAX)
-  {
+  if (output_length > COAP_PRINT_STATUS_MAX) {
     return COAP_PRINT_STATUS_ERROR;
   }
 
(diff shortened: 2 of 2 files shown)


tijuca:
Seems we have forgotten this PR.
Could you please rework the subject line and the message body a bit?

Please prefix the subject line instead of a simple "Foo fix". I suggest to do a subject like `net.c: correct brackets in if-statement`. And please start your commit message in line three of the body. Note the suggestions in [CONTRIBUTE chapter 5](https://github.com/obgm/libcoap/blob/develop/CONTRIBUTE#L161).

PawelWMS:
@tijuca: Sure, I can change the subject and the message to fit the guidelines. I've got a few questions for the future, though:
1. The instructions are not clear what to do with commits changing more than one file. Should we just pick one of the files we consider to be the best representation of the change?
2. From what I can see the commit message's body already is on the third line of the commit message. Have I missed something?

obgm:
The idea was to change the first line of the commit message, not the PR's title.

tijuca:
@PawelWMS 

Like Olaf said, the name of the PR doesn't matter. But the commit message is the thing that count as this is visibly in the history of the branches. Thanks!

PawelWMS:
Got it. I just always though that you can rephrase the merge commit when a PR is approved and considered it less invasive than force-pushing a commit, but I haven't done any PR merging in a while.

Was there also another way? Would pushing another commit in addition to the original one help? I'm trying to figure out the source of the commit message, which would end up in your repo after the PR is approved.

tijuca:
> I just always though that you can rephrase the merge commit when a PR is approved and considered it less invasive than force-pushing a commit

As far as I remember that isn't possible. That would be possible of course by the classical way posting the series through the ML and the commits being amended for such small changes.

You can simply force push the changes to your branch that's selected for the PR, the Travis hook is smart enough to being re-triggered. Here is a force push is the only way to solve this PR to get accepted. You could close this PR by yourself and reopen a new one, but this makes the idea behind the PR work flow a bit obsolete.

obgm:
I could indeed edit the commit message when doing the merge on the command line. I did that before for minor edits such as fixing spelling errors. But this has a few drawbacks. The most severe is that the merge history changes in a way that GH wont automatically display this PR as being merged.  Thus, I have begun avoiding even minor manual edits.

PawelWMS:
Got it. I've already force-pushed the comment change on Friday, though. Is there still something missing?

tijuca:
@obgm 
Could we just pick up this small change even if it needs local rebasing? The commit and changes are rather small and this PR is around for some time.

obgm:
Done. Thank you for picking this up.

PR #78, opened 2017-05-06
Support for windows platform

jcmichelou:
Add support for the windows (win32 with visual studio 2015) platform.
This PR incorporates as much of PR #47 as possible, after a lot of line by line cherry picking. It should be noted however that PR #47 **builds** on windows but is not functional, for example it does not cover receiving packets over UDP.
Changes in PR #47 related to TCP transport have been intentionally left aside.
The other notable deviation is that WITH_POSIX is no longer required and coap_config.h is not required to be included as a public header.
The included solution files build the static library, the examples and the regression test. To help building the regression test, a binary distribution of CUnit is included (in the win32 platform directory only).


obgm:
Okay, now I have a VS2017 up and running. It seems that I have to "retarget the toolset" (huh?) before it even starts trying to build the library. When I do this, it starts building uri.c for "Debug DLL x64" and then complains "fatal error C1083: Cannot open include file: 'coap_config.h': No such file or directory.
So, what is the VS-equivalent of configure?

(Sorry for the dumb newbie questions -- these IDE things just hide too much information about what is really going on...)

obgm:
The subdirectory `win32/coap-rp` should be renamed to `win32/coap-rd`, no?

jcmichelou:
The project files are for Visual Studio 2015. VS2017 is very widespread yet. It is possible to open VS2015 project files in VS2017 as is, they will appear with the mention (2015 project) next to them in the solution or to upgrade them to use the new toolchain of VS2017. Either should work.
I have only set the build configuration for Win32, so please go to build > configuration manager ... and choose "x86" as the "Active solution platform". This should switch all the lines in the "Platform" column to Win32.

jcmichelou:
Good catch for the typo with coap-rd directory name. I will modify this and submit it as new change in the pull request. I will try with VS2017 as well. In the meantime you should still be able to build and test coap-rd.

obgm:
Thanks, I have managed to convince VS2017 to build the executables in `win32\Debug DLL`. Apparently, all executables are suffixed with a letter `d` before the `.exe`. Why is that?

```
coap-clientd.exe
coap-rdd.exe
coap-serverd.exe
testdriverd.exe
```

And: I would have expected for `testdriver` either to fail because of an unsupported CUnit version or to work correctly. However, calling `testdriverd.exe` just returns (with `$?` being `False`).

jcmichelou:
To avoid confusion, especially with DLLs, I always add a 'd' suffix to librairies and executables from a debug build. Release builds don't have the suffix. This is a personal convention. Let me know if you want the suffix removed. By the way x64 builds should work as well now with df86b74, at least I have tested with coap-client and coap-server.

jcmichelou:
Did you use the CUnit DLL built with VS2015 that I have included in the pull request ? You might have to change the PATH to make sure which one it's using or just copy the dll next to the executable. If you use one built with mingw you must make sure it was linked with the system's C library as a DLL (msvcrt.dll) to be interoperable. In VS2017, you can make testdriver the "startup project", modify the path in the "debug" options of the project if needed and run with F5 it to see the list of DLLs loaded in the "modules" tab.

obgm:
Regarding the `d`-Suffix for debug vs. release builds: Okay, this makes sense to me.

jcmichelou:
Did you succeed in running testdriver ?
Here is how I run in it from bash in Debug DLL directory:

PATH=../CUnit:$PATH ./testdriverd.exe

And the end of the output:

Run Summary:    Type  Total    Ran Passed Failed Inactive
              suites     11     11    n/a      0        0
               tests    109    109    109      0        0
             asserts    852    852    852      0      n/a

Elapsed time =    0.012 seconds

Let me know if you have differing results.

obgm:
Thanks for the hint -- no this has worked as well. I also succeeded in building the library and client/server examples for x64. Apart from a couple of warnings that we can clear up after merging this PR (PR #47 already has proper fixes for them), this looks good so far.
One thing I am not happy with is integrating CUnit. Besides making license-related questions more complex, we would need at least have to provide binaries for both, x86 and x64 platforms. Therefore, I would like to go the same path as we do for Unix platforms and let users use their local installation of CUnit.

jcmichelou:
I fixed the two warnings in libcoap. The more numerous 64 bits / 32 bits size mismatch warnings in coap-client are not platform specific, they are also harmless as far as I can tell.

jcmichelou:
I submitted a 64-bit build of CUnit if you want to test with that.
I don't see any licensing problem with the redistribution of a binary build of CUnit since it's under the LGPLv2 license which allows precisely that. The user is free to substitute its own build of CUnit instead, it's just there for the convenience of having a Visual Studio solution file that you just press "Build solution" and it builds everything. Where it is a bit different from Linux is that there is no tool on Windows to pull a package with a pre-built binary distribution of CUnit and anybody willing to build testdriver would have to modify the solution to adjust the include path and link dependencies to make it point to its own copy of CUnit.
Anyways, CUnit comes with Visual C++ project files and all I have done is open the most recent one in VS2015 and do a build. For x64 it is a bit more complicated because the code in CUnit mixes up int and size_t in printf calls and those must be patched.
If you are uncomfortable with including CUnit, then I have no problem with having it removed either.

obgm:
I am in favor of removing CUnit.

jcmichelou:
OK. I removed CUnit.

Issue #68, opened 2017-03-13
coap_encode_var_bytes() invites buffer overflows


dmosberger:
Why not add a buf_len argument to that function so it can ensure it doesn't (unexpectedly) overwrite the caller's buffer?

tijuca:
Maybe you want to provide a patch?

dmosberger:
Sure, how about this?

https://github.com/obgm/libcoap/pull/79

tijuca:
Thanks for creating a patch with adding the extra functionality. But why create a new function with 90% of the same code? What can happen to existing binaries that will use the library with such a change?

Two possibilities, you create really a new function with the new check what will call the old function then e.g. after that check. Or you simply add the unique return code that will checked by the calling function, like you did in the new extra function in this patch, to the existing function.

I'd prefer to not add to much extra new wrapping function calls, especially like you named it. So I'd use option two.
If really a new function is needed a real clear and self-explanatory name is needed. Nobody can later really image what ...foo2 or ...foo3 really mean. I think it's better to improve the quality of existing code than to add new code with more or less the same code base.

But I'm not that deep in the source as Olaf, lets wait what he is thinking about your change.

dmosberger:
Sorry, I don't understand what you're saying here.

The patch maintains perfect backwards compatibility to existing binaries.

The idea of keeping the original version of the function is so that it can simply be removed once when it is decided that backwards compatibility is no longer needed (e.g., next major release).

Sure, you could re-write the old version of the function as:
```
unsigned int
coap_encode_var_bytes(unsigned char *buf, unsigned int val) {
  return coap_encode_var_bytes2(buf, ~0u, val);
}
```
If  you like that better, I can certainly update the patch.

Also, if it's OK to break source compatibility, you could add:

```
#define coap_encode_var_bytes(b,l,v) coap_encode_var_bytes2(b,l,v)
```
to ensure that newly compiled code breaks with a visible error when trying to use the old, unsafe interface.

If it's OK to use symbol versioning in libcoap, you could of course avoids having to introduce a new name by simply redirecting the old, unsafe interface to another name (e.g., old_coap_encode_var_bytes).

PR #79, opened 2017-05-12
Deprecate coap_encode_var_bytes() and add safer coap_encode_var_bytes2()

dmosberger:
coap_encode_var_bytes() will silently overwrite memory if the buffer
passed to it is too small.  coap_encode_var_bytes2() will instead
fails with a return value of 0.

### DIFF
--- a/include/coap/encode.h
+++ b/include/coap/encode.h
@@ -50,10 +50,22 @@ extern int coap_flsll(long long i);
 unsigned int coap_decode_var_bytes(unsigned char *buf,unsigned int len);
 
 /**
- * Encodes multiple-length byte sequences. buf points to an output buffer of
- * sufficient length to store the encoded bytes. val is the value to encode.
- * Returns the number of bytes used to encode val or 0 on error.
+ * DEPRECATED. Encodes multiple-length byte sequences. buf points to
+ * an output buffer of sufficient length to store the encoded
+ * bytes. val is the value to encode.  Returns the number of bytes
+ * used to encode val or 0 on error.
+ * Please use coap_encode_var_bytes2() instead.
  */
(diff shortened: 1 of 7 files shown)


mrdeep1:
I have fallen foul with not defining buf[] big enough several times now, so having this change would be useful for code safety.

I would like to see 4 changes though.

1) Delete the function coap_encode_var_bytes() in src/encode.c as it is really duplicated

2) Replace the function definition in include/coap/encode.h with
#define coap_encode_var_bytes(buf,val) \
    coap_encode_var_bytes2(uint8_t *buf, sizeof (buf), unsigned int val)

Worst case, if buf is a pointer rather than a uint8_t array, sizeof(buf) will be 4 (32 bit systems) or 8 (64 bit systems).  This will support backward compatibility if buf[] is a local array.

sizeof(buf) could be replaced with ~0u, then no checks would take place.  I think most existing usages of coap_encode_var_bytes() will have buf[] declared locally as an array. However, as unsigned int for parameter val will usually only be 4 bytes long, again, there should not be an issue. 

3) that an assert is used in coap_encode_var_bytes2() if there is not enough space (but still return 0 if assert() is disabled) to catch bad programming.
`    i >>= 8;

\+ assert(n > buf_len);
  if (n > buf_len)
    return 0;
`
4) Your code rebased to the current code as there are conflicting files.

tijuca:
@mrdeep1 
It's probably easier (given the age of this PR) if you pick up the original changes and adopt it with you suggested changes. Care to give it a try?

mrdeep1:
Makes sense to me

mrdeep1:
Fixed in #175.  Deprecated warning given if coap_encode_var_bytes() continues to be used.  #68 is also now answered by #175.

PR #84, opened 2017-05-25
Support for windows platform in dtls branch

jcmichelou:
This is a port to the dtls branch of the windows platform support.
I need both dtls and the windows platform, so I had to do this merge.
Clearly it is only meaningful If you first accept this pull request, then merge the dtls branch into develop, and forget about the PR for the develop branch. Both pull requests have basically the same content, but since the dtls branch is ahead of develop for network related stuff it'll be much easier to apply both in this order.
As it is now, there is no DTLS support for the windows platform yet. This requires a DTLS library. tinydtls could be ported for windows, though certainly not as is, but since others have requested OpenSSL support I think it might be the preferred library for cross-platform.


obgm:
Closing as this PR is rendered obsolete by the latest merge of changes from spinetix/dtls-dev.

PR #85, opened 2017-05-30
Generate .sym and .map

codable:
When I use libcoap, it's annoying that the .sym and .map files are updated after building and removed after distclean, both makes the repo dirty.
This PR remove .sym and .map from the repo and generates it when needed.


malinengineer:
If the files are auto-generated, then shouldn't they be added to .gitignore and that's it?

bukepo:
They are already added into the repo. And only generate when make dist before this PR.

malinengineer:
So why not remove them from the repo and add them to .gitignore..

codable:
I added them into .gitignore. They are not automatically generated with simply `make`, that's why I modified the Makefile.am to enable this.

obgm:
IIRC there was a reason why we do not generate these files with every build (except for that this requires an very specific version of ctags. As this file needs to be touched only when a) the API changes and b) a distribution is pacakged, I think that it is useful to have the latest stable version in the repository.
But I am not religious on this---@tijuca, what do you think?

tijuca:
I see there is still a issue with the distclean target, it's doesn't allowed to remove the map and sym file at all because users of a release tarball shouldn't be treated to regenerate those two files after a distclean call as this is needing additional software to recreate them like Olaf pointed. So the PR shouldn't be applied that way.

I'd like to not dynamically regenerate the two files as this shouldn't be need, even a dev wouldn't need to do that. Call update-map-file right after you've added or removed a symbol in the header files if you need new map and sym files. The package maintainer(s) are responsible for providing a full and complete tarball if a release is made. A developer should know what to do. Sorry but, I don't see the real point for changing the current behaviour in that way. The tarball need to be useable by the user even after distclean was called, I agree we need to adjust the distclean target a bit more. The main focus here need to be taken on the usability of tarball if we tunes something in the Makefiles and the environment and not if a developer has a local  impractical situation.

If my tree is dirty than this is o.k. if I know why I'm in a dirty state. BTW: The map and sym file are only recreated if you call `make dist` and this is what we wanted to archive by commits in the past.

codable:
1. Currently, these files are not maintained to be up to date.
2. With this PR, these files will not be generated every time you build. They are generated only when missing.
3. Even with this PR, these files will be put into tar ball, so it doesn't break the current behavior.

tijuca:
> Currently, these files are not maintained to be up to date.

From what do you assume that these files are not up2date? The libcoap library isn't in a heavy development state there a lot of symbols come and go, and if so the devs would make a bad job. Once again, in the end it's up to Olaf that at a release all active symbols are listed in booth files. From the old 4.1.1 to 4.1.2 finally only three symbols were added if I remember correctly.

> With this PR, these files will not be generated every time you build. They are generated only when missing.

You are really sure?
```
$ ./autogen.sh 
...
$ ./configure --disable-documentation
...
checking for doxygen... no
configure: WARNING: ==> Doxygen not found - continuing without Doxygen support
configure: WARNING: ==> The libcoap html documentation will not be build!
checking for ctags... HAVE_CTAGS=no
./configure: line 13440: HAVE_CTAGS=no: command not found
configure: ==> Note: Your ctags binary does not support '--c-kinds'!
configure: ==> Most likely, you are using the GNU Emacs ctag and not exuberant ctag.
configure: WARNING: ==> This option is required for the target 'update-map-file'.
configure: WARNING: ==> which is not a problem if you just want to build the library libcoap.
checking assert.h usability... yes
...
$ make
...
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -fPIC -I./include/coap -I./include/coap -pedantic -Wall -Wextra -Wformat-security -Winline -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wshadow -Wstrict-prototypes -Wswitch-default -Wswitch-enum -Wunused -Wlogical-op -Wunused-result -std=c99 -g -O2 -fdiagnostics-color -D_GNU_SOURCE -DWITH_POSIX -MT src/libcoap_1_la-uri.lo -MD -MP -MF src/.deps/libcoap_1_la-uri.Tpo -c src/uri.c -fPIE -o src/libcoap_1_la-uri.o >/dev/null 2>&1
mv -f src/.deps/libcoap_1_la-uri.Tpo src/.deps/libcoap_1_la-uri.Plo
( echo "VER_1 {" ; \
echo "global:" ; \
HAVE_CTAGS=no -I "coap_packet_extract_pbuf coap_pdu_from_pbuf" -f - --c-kinds=p ./include/coap/address.h ./include/coap/async.h ./include/coap/bits.h ./include/coap/block.h ./include/coap/coap.h ./include/coap/coap_io.h ./include/coap/coap_time.h ./include/coap/debug.h ./include/coap/encode.h ./include/coap/hashkey.h ./include/coap/libcoap.h ./include/coap/mem.h ./include/coap/net.h ./include/coap/option.h ./include/coap/pdu.h ./include/coap/prng.h ./include/coap/resource.h ./include/coap/str.h ./include/coap/subscribe.h ./include/coap/uri.h ./include/coap/uthash.h ./include/coap/utlist.h | awk '/^coap_/ { print "  " $1 ";" }' | sort ; \
echo "local:" ; \
echo "  *;" ; \
echo "};" ) > libcoap-1.map.new
/bin/bash: line 2: -I: command not found
mv libcoap-1.map.new libcoap-1.map
( HAVE_CTAGS=no -I "coap_packet_extract_pbuf coap_pdu_from_pbuf" -f - --c-kinds=p ./include/coap/address.h ./include/coap/async.h ./include/coap/bits.h ./include/coap/block.h ./include/coap/coap.h ./include/coap/coap_io.h ./include/coap/coap_time.h ./include/coap/debug.h ./include/coap/encode.h ./include/coap/hashkey.h ./include/coap/libcoap.h ./include/coap/mem.h ./include/coap/net.h ./include/coap/option.h ./include/coap/pdu.h ./include/coap/prng.h ./include/coap/resource.h ./include/coap/str.h ./include/coap/subscribe.h ./include/coap/uri.h ./include/coap/uthash.h ./include/coap/utlist.h | awk '/^coap_/ { print $1 }' | sort ) \
> libcoap-1.sym.new
/bin/bash: -I: command not found
mv libcoap-1.sym.new libcoap-1.sym
/bin/bash ./libtool  --tag=CC   --mode=link gcc -fPIC -fPIE -I./include/coap -I./include/coap -pedantic -Wall -Wextra -Wformat-security -Winline -Wmissing-declarations -Wmissing-prototypes -Wnested-externs -Wpointer-arith -Wshadow -Wstrict-prototypes -Wswitch-default -Wswitch-enum -Wunused  -Wlogical-op -Wunused-result -std=c99 -g -O2 -fdiagnostics-color -D_GNU_SOURCE -DWITH_POSIX -version-info 0:0:0 -Wl,--version-script=./libcoap-1.map -pie  -o libcoap-1.la -rpath /usr/local/lib src/libcoap_1_la-address.lo src/libcoap_1_la-async.lo src/libcoap_1_la-block.lo src/libcoap_1_la-coap_io.lo src/libcoap_1_la-coap_time.lo src/libcoap_1_la-debug.lo src/libcoap_1_la-encode.lo src/libcoap_1_la-hashkey.lo src/libcoap_1_la-mem.lo src/libcoap_1_la-net.lo src/libcoap_1_la-option.lo src/libcoap_1_la-pdu.lo src/libcoap_1_la-resource.lo src/libcoap_1_la-str.lo src/libcoap_1_la-subscribe.lo src/libcoap_1_la-uri.lo  
libtool: link: gcc -shared  -fPIC -DPIC  src/.libs/libcoap_1_la-address.o src/.libs/libcoap_1_la-async.o src/.libs/libcoap_1_la-block.o src/.libs/libcoap_1_la-coap_io.o src/.libs/libcoap_1_la-coap_time.o src/.libs/libcoap_1_la-debug.o src/.libs/libcoap_1_la-encode.o src/.libs/libcoap_1_la-hashkey.o src/.libs/libcoap_1_la-mem.o src/.libs/libcoap_1_la-net.o src/.libs/libcoap_1_la-option.o src/.libs/libcoap_1_la-pdu.o src/.libs/libcoap_1_la-resource.o src/.libs/libcoap_1_la-str.o src/.libs/libcoap_1_la-subscribe.o src/.libs/libcoap_1_la-uri.o    -g -O2 -Wl,--version-script=./libcoap-1.map   -Wl,-soname -Wl,libcoap-1.so.0 -o .libs/libcoap-1.so.0.0.0
/usr/bin/ld:./libcoap-1.map:3: syntax error in VERSION script
collect2: error: ld returned 1 exit status
Makefile:655: recipe for target 'libcoap-1.la' failed
```
Without a dependency on `$(lib_LTLIBRARIES)` the build went fine. The build here is breaking because I haven't installed exuberant ctag by default and the current constellation of configure.ac + Makefile.am isn't checking that more as wanted.

Hmm, the target check_ctags isn't working properly ... the issue above shouldn't happen.

> Even with this PR, these files will be put into tar ball, so it doesn't break the current behavior.

Yes, adding a automatic recreation of the map and sym file isn't breaking the actual behavior in relation of tarball creation. But adding such a behavior isn't needed, there is a existing target update-map-file that is called by Olaf or the contributor in preparation of a commit. I still don't see why we would need this kind of automatic. We should better work on some logic to check the completeness of a tarball if `make dist` is called. Currently there is nothing that helps to check that every thing is in place for a release.

obgm:
The symbol lists got dirty during build because a symbol (coap_flsll) from a previous PR was missing. After adding this, developers should not run into the issue described here.

Issue #83, opened 2017-05-24
Separate response handling


codable:
Request and response matching should use token instead of message id. However, the request is removed from sendqueue once an ACK of the same message id is received. This causes the real response is handled without the matching request.

obgm:
Yes, this was deliberately designed this way.

Keeping the initial request until the response is received needs to be done on the application layer.
A similar issue related to observe has been discussed on the mailing list some time ago [1]. Resolving the observe issue may produce a resolution to your problem with separate responses as wel.

[1] https://sourceforge.net/p/libcoap/mailman/message/35547934/

obgm:
Closing off as this issue has been fixed in PR #552

PR #86, opened 2017-05-30
Handle separate response

codable:
This PR aims at fixing #83. The theory is when receiving empty-message, it inserts the request into send queue and update the `retransmit_cnt` to be bigger than `COAP_DEFAULT_MAX_RETRANSMIT`, thus it will be dropped after the timer timeout, and the timeout is set be as if it is the last try of retransmission. When the actually response is received, the special request is removed immediately from the queue.


codable:
any comments?

obgm:
I haven't yet time to look into the patch, sorry. But will do soon.

src/net.c
```diff
@@ -1535,12 +1535,58 @@ handle_request(coap_context_t *context, coap_queue_t *node) {
...
+  }
+
+  p = context->sendqueue;
+  q = p->next;
+
+  /* when q is not NULL, it does not match (dst, token), so we can skip it */
+  while (q) {
+    if (coap_address_equals(dst, &context->sendqueue->remote) &&
+        token_match(token, token_length,
+                    context->sendqueue->pdu->hdr->token,
```
obgm: I presume this should be q instead of context->sendqueue, no?

src/net.c
```diff
@@ -1535,12 +1535,58 @@ handle_request(coap_context_t *context, coap_queue_t *node) {
...
+                    context->sendqueue->pdu->hdr->token_length) &&
+        context->sendqueue->retransmit_cnt > COAP_DEFAULT_MAX_RETRANSMIT) {
```
obgm: (see previous comment)

src/net.c
```diff
@@ -1535,12 +1535,58 @@ handle_request(coap_context_t *context, coap_queue_t *node) {
...
+			 const unsigned char *token, size_t token_length, coap_queue_t **sent) {
+  coap_queue_t *p, *q;
+  if (!context->sendqueue)
+    return 0;
+
+  if (coap_address_equals(dst, &context->sendqueue->remote) &&
+	    token_match(token, token_length,
+		              context->sendqueue->pdu->hdr->token,
+		              context->sendqueue->pdu->hdr->token_length) &&
+      context->sendqueue->retransmit_cnt > COAP_DEFAULT_MAX_RETRANSMIT) {
```
obgm: Maybe we can encapsulate the condition ...retransmit_cnt > COAP_DEFAULT_MAX_RETRANSMIT in some way so that the semantics are easier to understand, e.g.:

```c
static inline int
is_separate(const coap_queue_t *node) {
   return node->retransmit_cnt > COAP_DEFAULT_MAX_RETRANSMIT;
}
```
and then:
```c
...coap_address_equals(....) && token_match(....) && is_separate(context->sendqueue)...
```

src/net.c
```diff
@@ -1535,12 +1535,58 @@ handle_request(coap_context_t *context, coap_queue_t *node) {
...
+
 static inline void
 handle_response(coap_context_t *context, 
 		coap_queue_t *sent, coap_queue_t *rcvd) {
 
   coap_send_ack(context, &rcvd->local_if, &rcvd->remote, rcvd->pdu);
-  
+
+  /* Find separate response's request */
+  if (sent == NULL) {
```
obgm: It took me some time to understand why this will actually work---can you include a comment that explains why in this case, the sendqueue may contain the sent request despite `sent` being `NULL`?

codable:
Thanks for pointing out those mistakes and suggestions.

codable:
any more comments? @obgm

obgm:
Looks good at a first glance. I will do a quick test within the next days and merge on success,

src/net.c
```diff
@@ -1535,12 +1535,67 @@ handle_request(coap_context_t *context, coap_queue_t *node) {
...
   coap_send_ack(context, &rcvd->local_if, &rcvd->remote, rcvd->pdu);
-  
+
+  /* In case of separate response, the request cannot be matched with
+   * transaction id which is based on message id and remote address.
+   * Besides, excludes type Acknowledgement since separate response
+   * cannot be be of that type.
+   */
+  if (sent == NULL && rcvd->pdu->hdr->type != COAP_MESSAGE_ACK) {
+    coap_remove_separate_from_queue(context, &rcvd->remote,
```
obgm: `coap_remove_separate_from_queue()` removes a node from the send queue without deleting it afterwards. This results in a memory leak whenever a separate response is received before the maximum retransmission timeout is reached. You can check this e.g. with valgrind:

1. Start the example server: `coap-server`
2. From a different terminal, send a query for the async resource:  
     `valgrind -v coap-client -v9 -B 200 coap://[::1]/async?15`
3. (Optional:) Compare with very late response:  
     `valgrind -v coap-client -v9 -B 200 coap://[::1] /async?150`

An optimal solution to this should also avoid doubling most of the code in `coap_cancel_all_messages()`.

codable:
@obgm I updated the code to free the memory by not removing the separate request from the send queue before calling the actually handler and free them after calling handler. But I'm sorry, I cannot build with examples to verify memory leak as you suggested. The error is
```sh
a2x --doctype manpage --format manpage coap-client.txt
a2x: ERROR: "xmllint" --nonet --noout --valid "libcoap/examples/coap-client.xml" returned non-zero exit status 4
```

obgm:
Ugh, yes, saw that xmllint error before. IIRC, it has to do with a missing package that for some reason is not listed in the dependencies of a2x (some missing DTD or so). A quick and dirty workaround would be to pass `--disable-documentation` to `configure`.

codable:
I tested and it seems no memory leak found now.

codable:
any more comments?

PR #94, opened 2017-06-12
implement preliminary GnuTLS support

dbjorklu:
This pull adds the file coap_dtls_gnutls.c. It is not compiled (not added to the make system at all) so it will not affect anything.
It has been tested quite a bit and works, but there are things like some socket operations that should not be done in the dtls file I suppose (this only works in posix platforms)
keystore stuff has also not been implemented (easy to do it like in tinydtls file), a psk is currently hard coded in the file.


obgm:
Closing as this has been superseded by PR #273.

PR #100, opened 2017-08-08
Set relative time 't' to zero when it actually expire

jpcomtois:
Not seting that relative time to zero will only (in the pop process) increase the next node relative time with this one. 
e.g. if node1->t = 1000 (1 sec), and node2->t = 1000 ( + 1 sec),
after 1 sec, we pop node1 and node2 become first with t = 2000 (as pop will do node2->t += node1->t. But in reality, 1 second already elapsed, so node2->t should be 1000 instead.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1998,6 +1998,7 @@ static void coap_retransmittimer_execute(void *arg) {
       break;
     } else {
       elapsed -= nextinqueue->t;
+      nextinqueue->t = 0;
       coap_retransmit(ctx, coap_pop_next(ctx));
       nextinqueue = coap_peek_next(ctx);
     }


jpcomtois:
Travis CI build failed because develop is failing as well since a month ago.

tijuca:
> Travis CI build failed because develop is failing as well since a month ago.

Hmm, the GnuTLS package in trusty is unfortunately not recent enough
https://launchpad.net/ubuntu/trusty/amd64/gnutls-bin
that's why the Travis jobs are failing.

OTOH Trusty is equivalent to Debian Wheezy, which is while writing oldoldstable! Ubuntu Xenial has a version of gnutls-bin which can be used. I've no idea if Xenial can be used in a Travis setup. For testing within Debian we use docker containers with a free selectable release.

obgm:
The current behavior is correct as `node->t` is relative to `context->sendqueue_basetime`, no?

jpcomtois:
It is, in fact, normal to add the first node->t into the second, as they are relative. Although, As when the first node reaches the "retransmit", then node->t should become 0 (as we waited all the necessary times). What happens here, is that it is not reset to zero, so it keeps its previous node->t (in the example, 1 seconds), and then adjust all the next nodes with this. 

So what I said basically is that the behavior to distribute the node->t value into the next node is ok, but the problem is that this node->t value should be zero when it is. 

e.g, If the first node->t = 1000, we have to wait 1 sec before we retransmit.
After 1 seconds, the first node->t should equal zero right? But it actually keeps its previous value of 1000, and then after the retransmission, it distributes that value to the second (now first) node. If the "timer" goes off before the one seconds, let's say 500ms, then it'll update the node->t to 500, and restart the timer for 500ms. And when that timer pops, the current behavior will retransmit, and distribute that 500ms into the seconds' node (but it should be zero, as the time was up)

obgm:
You are right -- somehow I missed the magic happening in `coap_pop_next()`.
This very same fix should be necessary in `coap_write()` for platform POSIX. Oddly, there seems to be no difference when inserting `nextpdu->t = 0` before the call to `coap_pop_next()` in `coap_io.cc:730`...

PR #111, opened 2017-09-20
Assorted build fixes

diego-santacruz:
I encountered a few issues while building libcoap with DTLS support and I would like to push the fixes.

The first two concern building outside the source tree (for cleaner builds). 

The third one is to avoid exposing the SSL lib via pkg-config to libcoap users when not required as it can create problems in some scenarios.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,7 +146,7 @@ AX_CHECK_COMPILE_FLAG([-Wunused-result], [WARNING_CFLAGS="$WARNING_CFLAGS -Wunus
 
 AC_SUBST([WARNING_CFLAGS])
 
-AX_CHECK_LINK_FLAG([-Wl,--version-script=./libcoap-${LIBCOAP_API_VERSION}.map],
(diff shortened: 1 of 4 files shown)


obgm:
@tijuca, can you have a look at this as well? Since you have done the out-of-tree build, you might have an opinion on the removal of `-isystem`.

tijuca:
I'm fine with those changes but haven't done any checking on this. I expect @diego-santacruz has done a proper testing before raising a pull request. If we look further in the future we need to create some more test cases that will be run by Travis or a make target at least.

diego-santacruz:
I did test this in the context of a native Linux x64 build and also on a Yocto cross-compile and both work fine.

Diego Santa Cruz

Le 20 sept. 2017 à 19:35, tijuca <notifications@github.com<mailto:notifications@github.com>> a écrit :


I'm fine with those changes but haven't done any checking on this. I expect @diego-santacruz<https://github.com/diego-santacruz> has done a proper testing before raising a pull request. If we look further in the future we need to create some more test cases that will be run by Travis or a make target at least.

tijuca:
Yeah, then just go ahead.

Issue #90, opened 2017-06-03
Value of observe option repeated in the first 2 responses to a GET


jcmichelou:
If a client does a GET with an Observe option set to zero, the server must acknowledge the subscription for notifications by including an Observe option in the response (according to section 3.1 of RFC7641).
If ctx->observe is used as the value of the Observe option like it is in the server example then later on, when the resource becomes dirty, the client may receive a new GET response with an Observe option set to that same sequence number, but with the new content. This is because observe is incremented after the notification is sent, not before, so if no other notification is dispatched between the two events observe will have the same value. I suppose moving the incrementation of observe at the top of coap_notify_observer would fix this.

mrdeep1:
This is no longer an issue with the current code and can be closed off.

obgm:
Closing.

Issue #92, opened 2017-06-03
Value of observe option is global for all resources


jcmichelou:
Currently the value used for observe option in GET responses is a global variable common to all resources. This means it may increases more quickly than expected if you have many resources. Why not maintain an “observe” value per resource as suggested as a valid implementation in RFC ?
Rather than the public API being a modification of the internal state (dirty flag) in coap_resource_t, wouldn’t a coap_resource_set_dirty() function make sense ? It would be able to set the dirty flag without exposing it publicly, increment a per-resource observe sequence number which scales better in case the number of resources is large, and open the way to maintaining a linked list of dirty resources instead of having to scan through all the resources, including non-observable and non-observed ones in coap_check_notify().

obgm:
Closing as this has been fixed in PR #114

Issue #91, opened 2017-06-03
Observe option may exceed allowed size instead of wrapping


jcmichelou:
I have noticed that the “observe” variable in coap_context_t in used as a value for the Observe option in notification to clients. The maximum size of that option is 3 bytes, and RFC7641 section 4.4 is quite clear that the value is limited to 2^24-1 and should wrap. The way it is incremented currently, it could end up being greater or equal than 2^24, which is probably a bug.

obgm:
Closing as this has been fixed in PR #114

PR #114, opened 2017-10-25
Fix issues #90, #91 and #92

jcmichelou:
Changes in resource.c, resource.h:
The value of the observe option is kept for each resource rather than being global to the context.
The value of the observe option must wrap to be stay below 2^24.
The value of the observe option must be different from the initial value when a notification is sent.

Note that the dirty flag must no longer be touched by the application, coap_resource_set_dirty() must be called instead as it takes care of incrementing the observe value. The examples have been updated accordingly


obgm:
Thanks!

PR #128, opened 2018-01-26
Support for ECDSA use via tinydtls

bullekeup:
This PR propose a way to add support for ECDSA cipher use via tinydtls.
I tried to follow the design already used for PSK keys, with as little modifications as possible.

I added a dtls_handle pointer in struct coap_context_t, which allows using different contexts to work with or without ECDSA support in a per-context approach.

Keys are declared as in tinydtls, struct coap_dtls_ecdsa_key_t is a raw copy of the key declaration in tinydtls (struct dtls_ecdsa_key_t). They can be managed per session or per context, with ability to mix both. If a session provides a dedicated key, it is considered prioritary over context's ones (same way as PSK keys handling). Same applies for verify_ecdsa_key callback.

When a context / session starts using ECDSA, get_ecdsa_key and verify_ecdsa_key fields of dtls_handler_t are filled with actual callbacks and all further sessions needs to provide a dedicated key (or context's global key will be used if set) to avoid tinydtls errors when connecting to a client / server which supports ECDSA. That's why if a session use ECDSA, all sessions MUST support use of ECDSA, either via the context or per-session (and also why I choose to store the dtls_handler_t struct per-context).

I have made some changes to the configure.ac script to handle --without-ecc parameter (same as tinydtls), which unsets COAP_ECC_ENABLED define. When ECC is disabled, all my modifications will be ignored. It also allows users to check at compile time if libcoap was compiled with or without ECC support.

Still, this PR doesn't prevent an ECDSA enabled server / client from connecting from / to a PSK only server / client. In this case tinydtls will still use PSK keys provided.

I have tested theses changes with custom apps using both normal and block transfers, with  combinations of ECDSA enabled and ECDSA disabled clients and servers.


obgm:
Closed because this is being superseded by the recently merged PR #510.

PR #152, opened 2018-03-09
resource.h: Introduce new field user_data

baw-serafin:
The library API lacks a way to store context for the handlers. A
LED-handler for example needs to now on which pin the specific LED
is connected in order to control it. Or a higher level API would like to save the
object pointer to the object responsible for the resource.

This change will introduce a void pointer for the user to save context.
He can then retrive the context in the handler if needed.

### DIFF
--- a/include/coap/resource.h
+++ b/include/coap/resource.h
@@ -112,6 +112,12 @@ typedef struct coap_resource_t {
   */
   unsigned int observe;
 
+  /**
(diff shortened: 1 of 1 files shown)


baw-serafin:
Sorry did not realise the inline functions had doxygen too because the function I used as template used the short syntax. Did fix that

PR #156, opened 2018-03-14
Fire a Event on dtls error, when use tinydtls

cquijano:
Fire an event when there is an error on DTLS and we use tinydtls implementation.

### DIFF
--- a/src/coap_tinydtls.c
+++ b/src/coap_tinydtls.c
@@ -388,12 +388,15 @@ coap_dtls_receive(coap_session_t *session,
   size_t data_len
 ) {
   session_t *dtls_session = (session_t *)session->tls;
-  int res;
+  int err;
 
   coap_event_dtls = -1;
-  res = dtls_handle_message(
+  err = dtls_handle_message(
     (struct dtls_context_t *)session->context->dtls_context,
     dtls_session, (uint8 *)data, (int)data_len);
+
+  if (err)
+      coap_event_dtls = COAP_EVENT_DTLS_ERROR;
 
   if (coap_event_dtls >= 0) {
     coap_handle_event(session->context, coap_event_dtls, session);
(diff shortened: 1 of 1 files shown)


tijuca:
Could you please write a bit more explaining why you modified the source in the commit message?
https://github.com/obgm/libcoap/blob/develop/CONTRIBUTE#L165
I see you've added a if statement but I don't see what this is doing without digging into the code as you also don't have add a comment.

Also a prefix to the subject line is helpful to see if a commit is relevant while searching or reviewing.
https://github.com/obgm/libcoap/blob/develop/CONTRIBUTE#L179

src/coap_tinydtls.c
```diff
@@ -388,13 +388,16 @@ coap_dtls_receive(coap_session_t *session,
...
-  int res;
+  int err;
 
   coap_event_dtls = -1;
-  res = dtls_handle_message(
+  err = dtls_handle_message(
     (struct dtls_context_t *)session->context->dtls_context,
     dtls_session, (uint8 *)data, (int)data_len);
 
+  if (err)
```
obgm: Please use always `{` ... `}` for conditional blocks.

obgm:
@cquijano Do you have a version that addresses these comments?

cquijano:
@obgm not yet, I just forgot about this, let me check this, I Will re-do the pull request.
@tijuca , about the comment, you need to return an error in other to catch this error on an upper layer, this is the behavior with other implementations, not with tinydlts.

tijuca:
@cquijano 
I'm fine with this of course. But this information needs to go to the commit messages so other can see the reasoning behind your modification. Imagine you or others look t this commit in a year or so.

cquijano:
I am closing this, you have the new PR on #192.

PR #169, opened 2018-04-16
net.c: use after free in coap_free_context()

jcmichelou:
Because of incorrect free order, if there are open TLS connections when
coap_free_context() is called a use after free will occur as OpenSSL tries to
shutdown the connections.


jcmichelou:
Of course this causes a crash in this case, so this fix is high priority.

PR #170, opened 2018-04-18
removed un-necessary const cast

arabine:
It seems that this cast (which generates a build error with some compiler switches, at least a warning) is not necessary since the options are only accessed in read only. I enforced the original function with a 'const' argument passing.

It is the only "dangerous" code within the public interface, otherwise everithing is ok with all the compiler flags turned on. It would be good to have a very strict and compliant interface so that libcoap can be integrated in application environment with very strict compiler flags enabled.

### DIFF
--- a/include/coap/block.h
+++ b/include/coap/block.h
@@ -41,7 +41,7 @@ typedef struct {
  * returns @c NULL.
  */
 #define COAP_OPT_BLOCK_LAST(opt) \
-  (coap_opt_length(opt) ? (coap_opt_const_value(opt) + (coap_opt_length(opt)-1)) : 0)
+  (coap_opt_length(opt) ? (coap_opt_value(opt) + (coap_opt_length(opt)-1)) : 0)
 
 /** Returns the value of the More-bit of a Block option @p opt. */
 #define COAP_OPT_BLOCK_MORE(opt) \
--- a/include/coap/option.h
+++ b/include/coap/option.h
@@ -398,12 +398,7 @@ uint16_t coap_opt_length(const coap_opt_t *opt);
  *
  * @return    A pointer to the option value or @c NULL on error.
  */
-uint8_t *coap_opt_value(coap_opt_t *opt);
-
-COAP_STATIC_INLINE const uint8_t *
-coap_opt_const_value( const coap_opt_t *opt ) {
-  return coap_opt_value((coap_opt_t*)opt);
-}
+uint8_t *coap_opt_value(const coap_opt_t *opt);
 
 /** @} */
 
--- a/src/block.c
+++ b/src/block.c
@@ -31,9 +31,9 @@ coap_opt_block_num(const coap_opt_t *block_opt) {
   if (len == 0) {
     return 0;
   }
-  
+
   if (len > 1) {
-    num = coap_decode_var_bytes(coap_opt_const_value(block_opt), 
+    num = coap_decode_var_bytes(coap_opt_value(block_opt),
 				coap_opt_length(block_opt) - 1);
   }
   
--- a/src/option.c
+++ b/src/option.c
@@ -278,7 +278,7 @@ coap_opt_length(const coap_opt_t *opt) {
 }
 
 unsigned char *
-coap_opt_value(coap_opt_t *opt) {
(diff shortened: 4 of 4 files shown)


tijuca:
Could you please extend the commit message with the additional information from this PR? The PR message will not be visible later so it's always better to keep the commit message informative. Thanks!

mrdeep1:
I would like to see -Wcast-qual added to the WARNING_CFLAGS list in configure.ac to catch things like this earlier in the development cycle

tijuca:
@mrdeep1 
I'm fine with such additional CLAGS, and the intention was really to get a WARNING into a more strict handling. Sometime such flags a bit picky related to the compiler version, we can and should handle such case by a intelligent adding such flags related to the detected compiler version. I don't know if `-Wcast-qual` might be such a candidate, let see what @obgm is objecting about this.

mrdeep1:
I was not getting this as a warning as -Wcast-qual not one of -Wall or -Wextra in my compiler.
(Confirmed by https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)
I suggested it just as a warning rather than an error for the the reasons that you raised.

arabine:
tijuca > I can't find a way to do it! Any hint ?

tijuca:
Well, you probably just need to amend the last commit on your branch and do a force push afterwards.
```
git checkout removecast
git commit --amend
... [adjust the commit message]
git pull -f [your_remote] removecast
```
Travis will detect the changes and do a rebuild.

obgm:
I am happy to add `-Wcast-qual` to the list of warnings.

tijuca:
Hmm, why would this be needed? Why lower down the warning compiler flags for "outside" the core? I understand that we maybe need to distinguish the warning CFLAGS for the various platforms but currently I don't see why should have different flags over the code for the library.

arabine:
Because ofall the code that is considered as platform dependant. libcoap itself must be agnostic. Sometimes we have to comply to weird and exotic platforms, and in my experience we have to move everything that is not coap outside the library.Obgm  problem is hard to solve because you  cannot change window headers. None of my prototocols implementation have ifdefs, I have only (de)serializers. Only pure C code of buffers management.Le 20 avr. 2018 3:06 PM, tijuca <notifications@github.com> a écrit :Hmm, why would this be needed? Why lower down the warning compiler flags for "outside" the core? I understand that we maybe need to distinguish the warning CFLAGS for the various platforms but currently I don't see why should have different flags over the code for the library.

mrdeep1:
src/coap_io.c does not need a buffer copy, but a memcpy() replacment (safe as we are doing a send and so buffer will not be read into).  

\-    iov[0].iov_base = (uint8_t*)data;
\+    memcpy (&iov[0].iov_base, &data, sizeof (iov[0].iov_base));
... 
\-    mhdr.msg_name = (void *)&session->remote_addr.addr;
\+    memcpy (&mhdr.msg_name, &session->remote_addr.addr, sizeof (mhdr.msg_name));

src/debug.c is easily fixable by making coap_get_data() and coap_option_iterator_init() have pdu being const and removing an un-needed typecast.

Same probably is true for all the other files that are complaining

mrdeep1:
The code can easily be changed (in most cases by adding in const, a couple where const has to be removed) but there is a "design" challenge in 2 cases.

a) coap_resource_init() requires that the uri_path parameter is a const - and then has the potential flag entry of COAP_RESOURCE_FLAGS_RELEASE_URI to free it off when the resource is removed.
b) coap_add_attr() require that name and value are const values, but then has the potential flags entry of COAP_ATTR_FLAGS_RELEASE_NAME or COAP_ATTR_FLAGS_RELEASE_VALUE to free the appropriate value when the attribute is removed.

In both cases libcoap is trying to free off a const value if any of the release flags are set.

My suggestion would be to create 2 new functions coap_resource_init_release() and coap_add_attr_release() where the appropriate input parameters are not const and so are allowed to get freed off by libcoap when the appropriate object is removed.  What do you think?

mrdeep1:
Fixed in PR #173 now in develop.

obgm:
Closing as this was fixed as part of the extensive PR #173. Thanks again for starting this work.

PR #183, opened 2018-05-14
Monitor status of client sessions and send ping messages

jcmichelou:
Following my proposal on the mailing list, this PR implements new events so the application can track the status of client connections, should it want to recreate a session if it fails. For this purpose a new CSM exchange time out (default 30s) has been added. Automatic ping messages generation is optional, with a time out relative to when a message was last sent or received. The observer notification function has been fixed to work as documented for the return value.
As a side effect of introducing the new time outs there is also a fix to the server session garbage collection logic that was effectively not working as intended.


include/coap/net.h
```diff
@@ -279,6 +280,18 @@ int coap_context_set_psk( coap_context_t *context, const char *hint,
...
                            coap_dtls_pki_t* setup_data);
 
+/**
+ * Set the context keepalive timer for sessions.
+ * A keepalive message will be sent after if a session has been inactive,
+ * i.e. no packet sent or received, for the given number of seconds.
+ * For reliable protocols, a PING message will be sent. If a PONG has not
+ * been received before the next PING is due to be sent, the session will
+ * considered as disconnected.
+ * For unreliable protocols, an empty packet is sent
```
mrdeep1: Actually, the current changes do not do this for unreliable protocols.

mrdeep1:
Doing a code review, all looks to be good apart from the comment about the "ping" is not supported for DTLS.  I think it would be good to add this functionality in for DTLS.

Also, why does it need to be restricted to the client only?
- Agreed that it makes more sense for keeping client to server NAT translations warm than the other way around, but why the restriction

I have not had a chance to pull this code in and merge to see if anything else comes up.

include/coap/net.h
jcmichelou: Your are correct. I intended to do so initially, but then I realized it requires more changes since right now receiving a reset causes the client session to be shut down. I was also concerned that some implementations may as well clear the client state on their side when sending a reset.
I would propose to remove the line about unreliable protocols and push a separate pull request for the confirmable empty message / elicited reset sequence.

mrdeep1: Postponing doing the unreliable protocol and doing that as a separate PR works for me as it keeps the code changes separate.

jcmichelou:
Silence is more problematic for the client than for the server: in the case a client has subscribed to notifications, it must know if it is not receiving anything because of a network problem. On the server side, a broken connection will be detected by TCP when sending a notification. If there is no notification to send, then it does not really have any importance other than garbage collecting inactive sessions, which it does anyways. So in my opinion, servers do no need ping, as they will detect broken connections though TCP acknowledge timeouts.

include/coap/net.h
jcmichelou: OK, I fixed the documentation for the function in the header.

mrdeep1:
Agreed when using TCP, however when UDP is used, then this is a bit different and should be considered for that case.  For example https://tools.ietf.org/html/draft-ietf-dots-signal-channel-19#section-4.7 requires that both the server and client (the agents) "heartbeat" over both TCP (used as a backup) and UDP (primary traffic use), and is in https://tools.ietf.org/html/draft-ietf-dots-requirements-14#section-2.2 SIG-003 and SIG-004 requirements.

jcmichelou:
I don’t see any difficulty in the implementation. Do you think that the API should be selective, or just specify a common timeout for all client and server endpoints ? Or maybe the context-wide call should be for clients only and the timeout for servers should be set at the coap_endpoint_t level.
Since keep alive is based on an inactivity timeout, it would be beneficial to set the server-side timeout to a larger value than the one used by the client, so that the server does not need send pings if the client already does.

mrdeep1:
I think that ping_timeout  needs to be settable at the context level as you have in the current code, which then flows down to the session when it is instantiated. If this is to be used for DOTS, then it needs also to be configurable at the session level as the timeout value can change over time.

Yes, to minimize traffic, the server value should be set larger than that of the client.

jcmichelou:
OK, just rebased on develop @ 1cef78e

include/coap/coap_session.h
```diff
@@ -77,6 +77,7 @@ typedef struct coap_session_t {
   coap_tick_t last_rx_tx;
   coap_tick_t last_tx_rst;
   coap_tick_t last_ping;
+  coap_tick_t last_pong;
```
obgm: In the commit message, you state that the time accuracy is 1ms. This may or may not be true, depending on the system's clock resolution and the value of COAP_TICKS_PER_SECOND.

obgm: Sure, I understand. But I want to avoid that people read the commit message and get the impression that this 1ms resolution exists (and base assumptions on that).

jcmichelou: OK. Can you change the message during the merge or do you want me to amend the commit in the pull request ? The latter is not really push-button with git, unlike svn, but I guess I can do it by rebasing.

obgm: I think, I can change that manually (not push-button as well but not too hard).

examples/client.c
```diff
@@ -549,6 +553,7 @@ usage( const char *program, const char *version) {
      "\t       \t\tcommand line the same filename for both the certfile and\n"
      "\t       \t\tcafile (as in  '-c certfile -C certfile') to trigger\n"
      "\t       \t\tvalidation\n"
+     "\t-K interval\tsend a ping after interval seconds of inactivity\n"
```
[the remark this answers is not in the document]
mrdeep1: Furthermore, this should state for TCP sessions only.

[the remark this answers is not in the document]
obgm: I would rather suggest to implement the keep-alive mechanism for UDP as well, based on CoAP ping messages. (Maybe in another PR.)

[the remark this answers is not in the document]
jcmichelou: Agree UDP keepalive would be useful, but this need to be another PR as I won't have time to test and it's more delicate to implement as it uses packet types that are currently invalid.

obgm:
Closed after manual merge.

PR #192, opened 2018-05-31
coap_tinydtls.c: On DTLS error, fire and event that could be catch by upper layers

cquijano:
We need to fire an event on DTLS error, when we use tinydtsl, so we can handle this error on upper layers. 
On other implmetations of DTLS, like OpenSSL,this event is fired.
This is a fix of the PR #156


obgm:
In the commit message: s/could be catch/can be caught/.

cquijano:
@obgm I amend the last commit.

obgm:
Thank you!

cquijano:
@obgm You are welcome.

PR #195, opened 2018-06-02
fix typographical error in forward declaration

rdewaele:
This was the only match for a project-wide grep on 'coap_contex_t'.


obgm:
Thanks!

PR #211, opened 2018-06-27
oss-fuzz: Move test from oss-fuzz repo

bshastry:
Moves fuzzer test to source tree. Partly addresses #211

### DIFF
--- a/tests/oss-fuzz/pdu_parse_target.c
+++ b/tests/oss-fuzz/pdu_parse_target.c
@@ -1,0 +1,11 @@
+#include <coap.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
+{
+    coap_pdu_t *pdu = coap_pdu_init(0, 0, 0, size);
+    if (!pdu) return 0;
+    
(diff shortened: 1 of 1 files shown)


tests/oss-fuzz/pdu_parse_target.c
```diff
@@ -0,0 +1,11 @@
+#include <coap.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
```
obgm: I would prefer to follow the libcoap coding guidelines here, i.e., put the return type on a single line and
start the next line with the function identifier and end with the open bracket for the function block:

```c
int
LLVM....() {
```

tests/oss-fuzz/pdu_parse_target.c
```diff
@@ -0,0 +1,11 @@
...
+{
+    coap_pdu_t *pdu = coap_pdu_init(0, 0, 0, size);
+    if (!pdu) return 0;
```
obgm: In this case, I prefer:

```c
if (pdu) {
  coap_pdu_parse...
  coap_delete_pdu...
}
return 0;
```

PR #202, opened 2018-06-11
Various patches: naming and debug print

arabine:
This PR contains several patches not linked together. First I renamed various things to avoid conflict with other files ( _COAP_CONFIG_H_ is better than the generic _CONFIG_H_).

I added also some debug print which are useful to easily track bad defines.

PR #255, opened 2018-09-18
Corrected build for mem.c and resource.h

arabine:
Here are two patches that correct the build on my system:
  - in mem.c the string_t structure is already defined elsewhere, so we need to find another name. I don't know why you don't have the problem, maybe because of a different build order
 - In encode.h, there is a change of the sign for the deprecated coap_encore_var_bytes() which fails for restrictive build system environment

### DIFF
--- a/include/coap/encode.h
+++ b/include/coap/encode.h
@@ -87,10 +87,10 @@ unsigned int coap_encode_var_safe(uint8_t *buf,
  *   coap_encode_var_safe(buf, sizeof(buf), 0xfff);
  * would catch this error at run-time and should be used instead.
  */
-COAP_STATIC_INLINE COAP_DEPRECATED int
-coap_encode_var_bytes(uint8_t *buf, unsigned int value
-) {
-  return coap_encode_var_safe(buf, sizeof(value), value);
-}
+//COAP_STATIC_INLINE COAP_DEPRECATED int
+//coap_encode_var_bytes(uint8_t *buf, unsigned int value
+//) {
+//  return coap_encode_var_safe(buf, sizeof(value), value);
+//}
 
 #endif /* _COAP_ENCODE_H_ */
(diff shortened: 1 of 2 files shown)


mrdeep1:
I think the build order issue is down to you doing a Contiki build.

src/mem.c is corrected in PR #208 (in part based off your work in #202) by using a different name (coap_stringbuf_t), and it would be good if aligned with that change.  Even better it would be good for you to try and use the code in PR #208 for further testing. 

Apologies - a last minute change to coap_encode_var_safe()'s return value was not picked up for the deprecated coap_encode_var().

arabine:
Ok I did not understand why it was still not corrected on develop branch. Will your #208 PR be merged before 4.2 delivery?

mrdeep1:
That certainly is something I would like to see merged into 4.2

obgm:
Ok, will do.

(Sorry for stalling this for so long — still aiming for end of September 2018 for RC1.)

include/coap/encode.h
```diff
@@ -87,10 +87,10 @@ unsigned int coap_encode_var_safe(uint8_t *buf,
  *   coap_encode_var_safe(buf, sizeof(buf), 0xfff);
  * would catch this error at run-time and should be used instead.
  */
-COAP_STATIC_INLINE COAP_DEPRECATED int
-coap_encode_var_bytes(uint8_t *buf, unsigned int value
-) {
-  return coap_encode_var_safe(buf, sizeof(value), value);
-}
+//COAP_STATIC_INLINE COAP_DEPRECATED int
```
obgm: If you think this was the right time to remove that function, it should be deleted completely.
(Besides: please use always the block comment syntax.)

obgm:
I took the missing cast from commit 0c710327f and applied it as commit d0dfa0b. I suppose, this PR can be closed then.

PR #296, opened 2019-02-02
Update configure.ac

yr2004:
fix a configure.ac error for tinydtls


obgm:
Good catch, thank you!

PR #303, opened 2019-02-26
option.c: memset the node when a new optlist node is created

yr2004:
memset the node when a new optlist node is created, because the node->next maybe not NULL, so if it creates a new optlist node, and delete it, and then create it again, error will be occurred


obgm:
Good catch, thank you! Can you please prefix the commit message with the name of the changed source file like this: `option.c: memset the`...?

obgm:
Sorry, I meant to change the first line of the commit message (the PR's title is not that important).

obgm:
Nevermind, I have adapted that manually in commit e269b87

PR #318, opened 2019-03-28
coap_debug.c: fix format warning when compiling with GCC 8

igrr:
When compiling libcoap with format truncation warnings enabled,
GCC warns about the potential issue with "%u.%02u" format string,
since the first argument may produce more than one byte of output:

    error: '%02u' directive output may be truncated writing 2 bytes
    into a region of size between 0 and 3 [-Werror=format-truncation=]

The first argument (class) is limited to 3 bits according to the
specification. Mask the higher bits to fix the warning.

### DIFF
--- a/src/coap_debug.c
+++ b/src/coap_debug.c
@@ -291,7 +291,7 @@ msg_code_string(uint16_t c) {
   } else if (c >= 224 && c - 224 < (int)(sizeof(signals)/sizeof(const char *))) {
     return signals[c-224];
   } else {
-    snprintf(buf, sizeof(buf), "%u.%02u", c >> 5, c & 0x1f);
+    snprintf(buf, sizeof(buf), "%u.%02u", (c >> 5) & 0x7, c & 0x1f);
     return buf;
   }
 }


igrr:
Travis has failed with what seems to be an unrelated issue:

    tests/testdriver: error while loading shared libraries: libtinydtls.so: cannot open shared object file: No such file or directory

Looking at Travis build history, it looks like the failure has been introduced in https://github.com/obgm/libcoap/commit/641b5c174db6dc633a329c26642cd659358c9e97. Should I rebase the PR on the previous commit in `develop` branch (which had passed CI) or wait until CI is fixed in master?

mrdeep1:
@igrr There is no point in rebasing to something before https://github.com/obgm/libcoap/commit/641b5c174db6dc633a329c26642cd659358c9e97 other than to get it through the Travis build as any new PRs are added to the latest develop code.

The issue with the latest tinydtls adding in building libtinydtls.so as well as libtinydtls.a is confusing things as Travis is not doing a `make install` and so when running the test executables that are trying to load in  libtinydtls.so (even though --disable-shared is set) from the general library directory - and fails.  This needs to get fixed (as well as dropping the requirement for --disable-shared with TinyDTLS).

obgm:
Thanks!

PR #341, opened 2019-04-29
modify pud buf size

yr2004:
buf size should be larger than the received data, otherwise when printing data, it will print "hello worldXXX", not  "hello world", and buf should be memseted

### DIFF
--- a/src/pdu.c
+++ b/src/pdu.c
@@ -113,11 +113,12 @@ coap_pdu_init(uint8_t type, uint8_t code, uint16_t tid, size_t size) {
 #else /* WITH_LWIP */
   uint8_t *buf;
   pdu->alloc_size = min(size, 256);
-  buf = coap_malloc_type(COAP_PDU_BUF, pdu->alloc_size + pdu->max_hdr_size);
+  buf = coap_malloc_type(COAP_PDU_BUF, pdu->alloc_size + pdu->max_hdr_size + 1);
   if (buf == NULL) {
     coap_free_type(COAP_PDU, pdu);
     return NULL;
   }
+  memset(buf, 0, pdu->alloc_size + pdu->max_hdr_size + 1);
   pdu->token = buf + pdu->max_hdr_size;
 #endif /* WITH_LWIP */
   coap_pdu_clear(pdu, size);


mrdeep1:
There is a dangerous assumption here that the data variable inside coap_pdu_t is always ascii printable (but actually is typed as uint8_t).

You do not state where you are doing the printing of data, but whenever you print off any internal (to libcoap) data that has a length associated with it, it should be done as

printf(".*s", ((int)length, buffer).

Furthermore, zeroing data has as CPU and more importantly a write memory access (causing cache flushes) overhead which can add significant delays if used frequently.  This should be done where only necessary.

yr2004:
1/ "hello world" is just a example
2/ such as the func message_hanler in client.c, in this func, the "received" not has data length of data, so if the user want to get data len, he will use strlen(data), but unfortunatly, strlen(data) maybe a wrong value

mrdeep1:
There are several things here that need to be considered.

The type of data received in a packet cannot be assumed to ascii printable - there may even be embedded NULs in the data.  Only if the COAP_OPTION_CONTENT_TYPE is COAP_MEDIATYPE_TEXT_PLAIN is it reasonably safe to assume that it is 7bit printable ascii which the likes of strlen() can safely work on - see https://tools.ietf.org/html/rfc7252#section-5.10.3 for further information about content types.

In client.c, in message_handler(), there is the following line
````
if (coap_get_data(received, &len, &databuf))
````
which abstracts the data pointer and length of data from the received PDU information.  this `len `must be used instead of strlen().

In general, you should not access the libcoap internal structures from within an application, but should try to get to the information through the various abstract functions (e.g. coap_get_data() above).

I agree that it is frustrating that ascii readable string are not passed around everywhere, but, for example, char* is not the same (not always a signed byte pointer) on different platforms which can confuse things, so the decision was to go for uint8_t* and lengths for all the data components.

yr2004:
I agree with u, there is no harm in doing so.  i have another question, libcoap can not support multiple clients, maybe u can consider it

mrdeep1:
All new questions should be discussed under a separate thread to make tracking of conversations easier.

However, libcoap supports multiple coap_context_t* (the application needs to manage them), as well as a single context can be used for setting up multiple client sessions which works with PKI and no-encryption today - with #335 merged into develop, that will support PSK as well.

yr2004:
1、support multi coap server at the same time, such as coap.me/AWS iot

mrdeep1:
> 1、support multi coap server at the same time, such as coap.me/AWS iot

I'm sorry - I do not follow what you mean in your comment.  A CoAP server using libcoap can support many active clients, each potentially with different authentication keys.

obgm:
Closing this as I do not think that the buffer size is an issue. CoAP is supposed to carry binary data so you would have to control what is printed using the length anyway.

As to zeroing out the buffer, I would do this in `coap_pdu_clear()`. But I am still not convinced that this is really necessary.

PR #342, opened 2019-04-29
Update net.c

yr2004:
the packet buf should be memseted


mrdeep1:
Is there a specific reason as to why you need to clear s_packet before populating (using packet) the information within s_packet?

obgm:
?

PR #344, opened 2019-04-29
modify node buf size when it is malloced

yr2004:
the node buf should be larger than the data len

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -554,10 +554,10 @@ coap_new_optlist(uint16_t number,
 ) {
   coap_optlist_t *node;
 
-  node = coap_malloc_type(COAP_OPTLIST, sizeof(coap_optlist_t) + length);
+  node = coap_malloc_type(COAP_OPTLIST, sizeof(coap_optlist_t) + length + 1);
 
   if (node) {
-    memset(node, 0, (sizeof(coap_optlist_t) + length));
+    memset(node, 0, (sizeof(coap_optlist_t) + length + 1));
     node->number = number;
     node->length = length;
     node->data = (uint8_t *)&node[1];


mrdeep1:
I am confused as to why this is needed.

yr2004:
maybe u not check the "optList" in client.c, if u have look at it, u will find that, the string of optList->data is not end with '0', so it looks like "uri-pathXXX", not "uri-path", it looks very  strange

mrdeep1:
OK - so are you saying that you need this to help your with your debugging?
It is not needed for general code operation that uses data defined by` uint8_t* buffer, sizet length`.
[By the way I do not have a optList (upper L) variable in client.c, so that may be specific to your code]

It should also be noted that URI_PATH can contain data that is not ascii printable characters.
https://tools.ietf.org/html/rfc7252#section-6.5 bullet 6 refers to how any data you provide in the optlist for Uri-Path is converted into what is actually put into the packet that is transmitted.  The reverse is done in https://tools.ietf.org/html/rfc7252#section-6.4 bullet 8.  At no point is this restricted to ascii data and furthermore explicitly covers how to handle bytes  that are not in the 'allow unchanged set".  You can conveniently use ascii here, but there is no guarantee that the other end of the connection will do the same.

yr2004:
1/ the file is in examples/client.c, line 46, i modify my client base on it to support multi clients
2/ uri-path is just a example
3/ the change only helps to debug

mrdeep1:
1/ OK - you changed the variable name from optlist (lower L) to optList (upper L) on line 46.
2/ The same will be true for every option and every data type (the data could be binary with a specific length).
3/ If you are debugging by using coap_log() debug statements, then you can use `coap_log(LOG_DEBUG, "The buffer is .*s", (int)len, buf);`  If using gdb it gets a bit more difficult which is maybe what you are seeing.

yr2004:
1、i porting libcoap to freertos with POSIX, and i debug it with keil

obgm:
Closing this PR as I do not see why it is needed.

PR #350, opened 2019-05-28
Fixed two minor bugs in endpoint free and session address update

lewixliu:
When try to use this library, I found two minor bugs.

1. Sometimes the session remote_addr and local_addr are not updated from the packet received. I just found the address copy function made the wrong direction.
2. Remove the endpoint from context list when free'd in coap_free_endpoint().


mrdeep1:
Copy Address direction fix.

As coap_packet_set_addr() is only used in one place, I would replace that caller with the corrected 2 lines and then remove all references to coap_packet_set_addr().  This make the code more understandable and the fix more obvious.

PR #404, opened 2019-09-10
Pthread configure check

obgm:
This PR proposes a configure-time check for pthread mutexes as discussed in PR #401 which would then be obsolete.


mrdeep1:
Yes, this looks a lot better than #401.

obgm:
Ok, thanks. So I will merge this one.

PR #412, opened 2019-09-23
libcoap.vxproj: Fix hard-wired PlatformToolset

obgm:
Use DefaultPlatformToolset instead of hard-wired values as suggested
in [1].

[1] https://www.appveyor.com/docs/lang/cpp/


mrdeep1:
This also needs to be done in the other 5 .vcxproj files.  Certainly VS2019 is not complaining with this change.  This does not handle the TargetPlatformVersion though

obgm:
Closing in favor of the now more complete #411.

PR #415, opened 2019-10-19
Fix epoll detection

cgudrian:
The "ac_cv_header-..." macros are either set to "yes" or "no" and never to
the empty string.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -575,7 +575,7 @@ AC_CHECK_HEADERS([assert.h arpa/inet.h limits.h netdb.h netinet/in.h \
 # For epoll, need two headers (sys/epoll.h sys/timerfd.h), but set up one #define
 AC_CHECK_HEADER([sys/epoll.h])
 AC_CHECK_HEADER([sys/timerfd.h])
-if test "x$ac_cv_header_sys_epoll_h" != "x" -a "x$ac_cv_header_sys_timerfd_h" != "x"; then
+if test "x$ac_cv_header_sys_epoll_h" = "xyes" -a "x$ac_cv_header_sys_timerfd_h" = "xyes"; then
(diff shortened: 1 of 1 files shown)


obgm:
Thank you! Please prefix the first line of the commit message with `configure.ac: ` as described in the contribution guidelines.

cgudrian:
Done.

configure.ac
```diff
@@ -575,7 +575,7 @@ AC_CHECK_HEADERS([assert.h arpa/inet.h limits.h netdb.h netinet/in.h \
 # For epoll, need two headers (sys/epoll.h sys/timerfd.h), but set up one #define
 AC_CHECK_HEADER([sys/epoll.h])
 AC_CHECK_HEADER([sys/timerfd.h])
-if test "x$ac_cv_header_sys_epoll_h" != "x" -a "x$ac_cv_header_sys_timerfd_h" != "x"; then
+if test "x$ac_cv_header_sys_epoll_h" == "xyes" -a "x$ac_cv_header_sys_timerfd_h" == "xyes"; then
```
obgm: `==` is a bashism, please use `=` instead: 

```shell
sh -c 'test "x$ac_cv_header_sys_epoll_h" == "xyes"'
sh: 1: test: x: unexpected operator
```

cgudrian: Interestingly the above code works on my system (macOS). From the manpage:
> For compatibility with some other implementations, the = primary can be substituted with == with the same meaning.
I've adjusted the code nonetheless, of course.

obgm:
Uh, I have clicked "merge" too fast: In the future, when fixing one issue after review (in this case properly prefixing the commit line), **please, please, please do not revert things that have already fixed with force-pushes**!

PR #426, opened 2019-12-03
bugfix: ea72c3d7 outstanding CON requests limitation may drop new request.

ltaskpt:
The r->partiallydirty is set to 0 before the limitation. A new
request will be dropped when last request hasn't receive it's ACK.
Restore r->partiallydirty to 1 in the condition.

### DIFF
--- a/src/resource.c
+++ b/src/resource.c
@@ -760,8 +760,11 @@ coap_notify_observers(coap_context_t *context, coap_resource_t *r) {
         continue;
       if (obs->session->con_active >= COAP_DEFAULT_NSTART &&
           ((r->flags & COAP_RESOURCE_FLAGS_NOTIFY_CON) ||
-           (obs->non_cnt >= COAP_OBS_MAX_NON)))
+           (obs->non_cnt >= COAP_OBS_MAX_NON))) {
+        r->partiallydirty = 1;
+        obs->dirty = 1;
         continue;
+      }
 
       coap_tid_t tid = COAP_INVALID_TID;
       obs->dirty = 0;


obgm:
Thank you. Could you please prefix the commit message with `resource.c: `. The commit number that introduced the issue can go into the commit message. 
In your commit message: 
s/when last request/when the last request/
s/receive/received/.

ltaskpt:
> Thank you. Could you please prefix the commit message with `resource.c: `. The commit number that introduced the issue can go into the commit message.
> In your commit message:
> s/when last request/when the last request/
> s/receive/received/.

ok～

obgm:
Thanks. Unfortunately, I have missed a typo: s/it's ACK/its ACK/

ltaskpt:
> Thanks. Unfortunately, I have missed a typo: s/it's ACK/its ACK/

done～

ltaskpt:
> Thanks. Unfortunately, I have missed a typo: s/it's ACK/its ACK/
I changed my commit，because the force clear operation of 'r->dirty' after 'LL_FOREACH' loop will
drop new request too.

obgm:
Looks good. @mrdeep1 any objections?

mrdeep1:
`return; ` change needs to be replaced by `continue;`  Reviewing the dirty/partialdirty logic, this fix otherwise looks good.

mrdeep1:
@ltaskpt @obgm The following occurred to me while out today

It needs to be more than just changing `return; `to `continue;` - `obs->dirty` needs to be set (as only `r->dirty` may have been set) and so `r->partialdirty` needs to be set as well.  So the complete patch should be
````
diff --git a/src/resource.c b/src/resource.c
index 5644212..c3d491d 100644
--- a/src/resource.c
+++ b/src/resource.c
@@ -760,8 +760,11 @@ coap_notify_observers(coap_context_t *context, coap_resource_t *r) {
         continue;
       if (obs->session->con_active >= COAP_DEFAULT_NSTART &&
           ((r->flags & COAP_RESOURCE_FLAGS_NOTIFY_CON) ||
-           (obs->non_cnt >= COAP_OBS_MAX_NON)))
+           (obs->non_cnt >= COAP_OBS_MAX_NON))) {
+        obs->dirty = 1;
+        r->partiallydirty = 1;
         continue;
+      }
 
       coap_tid_t tid = COAP_INVALID_TID;
       obs->dirty = 0;
````
This also follows the method used for subsequent set ups for retrying the output in coap_notify_observers().

@ltaskpt Please confirm that this works in the environment where you found the issue.

ltaskpt:
> @ltaskpt @obgm The following occurred to me while out today
> 
> It needs to be more than just changing `return; `to `continue;` - `obs->dirty` needs to be set (as only `r->dirty` may have been set) and so `r->partialdirty` needs to be set as well. So the complete patch should be
> 
> ```
> diff --git a/src/resource.c b/src/resource.c
> index 5644212..c3d491d 100644
> --- a/src/resource.c
> +++ b/src/resource.c
> @@ -760,8 +760,11 @@ coap_notify_observers(coap_context_t *context, coap_resource_t *r) {
>          continue;
>        if (obs->session->con_active >= COAP_DEFAULT_NSTART &&
>            ((r->flags & COAP_RESOURCE_FLAGS_NOTIFY_CON) ||
> -           (obs->non_cnt >= COAP_OBS_MAX_NON)))
> +           (obs->non_cnt >= COAP_OBS_MAX_NON))) {
> +        obs->dirty = 1;
> +        r->partiallydirty = 1;
>          continue;
> +      }
>  
>        coap_tid_t tid = COAP_INVALID_TID;
>        obs->dirty = 0;
> ```
> 
> This also follows the method used for subsequent set ups for retrying the output in coap_notify_observers().
> 
> @ltaskpt Please confirm that this works in the environment where you found the issue.

The reason I change 'continue' to 'return' is that, if there are more than two notifications(different resources) at the same time, coap_notify_observers() will handle the first one and wait for the ACK, the second one will be dropped because 'r->dirty' of the second one will be set to zero at the end of loop.

mrdeep1:
@ltaskpt Thanks for checking and pushing the updated code.

obgm:
Thank you @ltaskpt and @mrdeep1!

PR #455, opened 2020-02-18
OpenSSL: Support self-signed certificates at depth 0

biodranik:
When libcoap config allows self-signed certificates (`coap_dtls_pki_t.allow_self_signed == 1`) the library should also accept self-signed certificates at depth 0. This is achieved by allowing [X509_V_ERR_DEPTH_ZERO_SELF_SIGNED_CERT](https://www.openssl.org/docs/man1.1.0/man3/X509_STORE_CTX_get_error_depth.html) case.

### DIFF
--- a/src/coap_openssl.c
+++ b/src/coap_openssl.c
@@ -1418,6 +1418,7 @@ tls_verify_call_back(int preverify_ok, X509_STORE_CTX *ctx) {
       if (setup_data->allow_expired_certs)
         preverify_ok = 1;
       break;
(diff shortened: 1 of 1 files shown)


mrdeep1:
Change looks good to me.

biodranik:
@mrdeep1 who is responsible for the merge?

mrdeep1:
Merge responsibility is obgm.

obgm:
Here we go...

PR #474, opened 2020-03-12
Coap sessions not closed if observing clients

martinb001:
Coap sessions are not always closed when deleting
a resource. If there are clients that observe a specific
resource, then when deleting the resource will not close
the active session.

This fix "forces" coap sessions to close when a resource is
deliberately removed.

Change-Id: Ib0a0503647fdf14991169553adf22306c9c59857

### DIFF
--- a/include/coap2/coap_session.h
+++ b/include/coap2/coap_session.h
@@ -134,10 +134,13 @@ coap_session_t *coap_session_reference(coap_session_t *session);
 * Decrement reference counter on a session.
 * Note that the session may be deleted as a result and should not be used
 * after this call.
-*
+* The v2 version was created as a "bugfix" when the associated coap sessions
+* were not fully released when deleting a resource. This fix is more a workaround
+* that a proper solution.
 * @param session The CoAP session.
 */
 void coap_session_release(coap_session_t *session);
+void coap_session_release_v2(coap_session_t *session);
 
 /**
 * Stores @p data with the given session. This function overwrites any value
--- a/src/coap_session.c
+++ b/src/coap_session.c
@@ -70,6 +70,7 @@ coap_session_reference(coap_session_t *session) {
   return session;
 }
 
+
 void
 coap_session_release(coap_session_t *session) {
   if (session) {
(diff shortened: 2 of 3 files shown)


mrdeep1:
I'm a bit confused as to why removing the session actually is needed and it has the potential to break things such as (D)TLS sessions.

Removing the resource should delete all observers of that resource - and so the client will stop seeing any observe update traffic.  Any subsequent (GET) attempt to access the resource by the client should then return 4.04.

The current design is that the listening endpoint on the CoAP server "accepts" (by creating a coap_session_t) new incoming sessions (which may set up a (D)TLS based session).  The session is then left there for a period of time before it is deleted (which closes down the (D)TLS stuff if appropriate) should the next request (which can be to a different resource) comes in from the client using the same src ip/port/protocol.  If (D)TLS information has gone, then (D)TLS will have to be set up / resumed which the client will need to initiate.  Your solution will exacerbate this restart, unless I am missing something.

martinb001:
I think I see your point. There may be a penalty reestablishing the DTLS connection if we do as I suggest. I agree that my suggested solution may not be the best.

However, this is what the RFC says (4.2.  Notifications):
   ...
   A notification SHOULD have a 2.05 (Content) or 2.03 (Valid) response
   code.  However, in the event that the state of a resource changes in
   a way that would cause a normal GET request at that time to return a
   non-2.xx response (for example, when the resource is deleted), the
   server SHOULD notify the client by sending a notification with an
   appropriate response code (such as 4.04 Not Found) and subsequently
   MUST remove the associated entry from the list of observers of the
   resource.

mrdeep1:
A good point.  Currently libcoap does not send the 4.04 notification as per
````
server SHOULD notify the client by sending a notification with an
appropriate response code (such as 4.04 Not Found)
````
when deleting the resource , but does remove the associated entry from the list of observers of the resource.

It is possible to get a 4.04 sent out when a resource is deleted - need to be careful in the coding though to use as much as possible of the existing functionality.

Would this help with your original issue?

martinb001:
Yes, I believe it would help. It also depends a little on what the coap-client (example/client.c) would do on such response. I would like it to terminate. Took a quick look at the code and looks like it doesn't.

PR #557, opened 2020-10-20
coap_mbedtls.c: verify certificate after successful handshake

obgm:
Peter van der Stok has pointed out that according to the documentation,
the server certificates should be verified when peer verification is not
used. As a resolution, he has created this fix.

Signed-off-by: Peter van der Stok <consultancy@vanderstok.org>

### DIFF
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -1109,10 +1109,24 @@ coap_dtls_free_mbedtls_env(coap_mbedtls_env_t *m_env) {
 static int do_mbedtls_handshake(coap_session_t *c_session,
                                 coap_mbedtls_env_t *m_env) {
   int ret;
+  uint32_t flags = 0;
 
   ret = mbedtls_ssl_handshake(&m_env->ssl);
   switch (ret) {
-  case 0:
+  case 0: /* success */
+    /* try certificate verification */
+    flags = mbedtls_ssl_get_verify_result(&(m_env->ssl));
+    if (flags != 0) {
+      char vrfy_buf[COAP_DEBUG_BUF_SIZE];
+      int length;
+      length = mbedtls_x509_crt_verify_info(vrfy_buf,
+                                            sizeof(vrfy_buf),
+                                            " ! ",
+                                            flags);
+      coap_log(LOG_WARNING, "certificate verification failed: %*.s\n",
+               length, vrfy_buf);
+    }
+
     m_env->established = 1;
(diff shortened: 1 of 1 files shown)


obgm:
@mrdeep1 can you have a look at this proposal? The issue was that OpenSSL and MbedTLS showed a different behavior when `require_peer_cert` was set to 1.

mrdeep1:
Sure

mrdeep1:
Following offline discussions with Peter it transpires that if the coap-client and coap-server use different self-signed certificates (i.e there is not  common CA signing both the certificates so mutual authentication cannot take place at the (D)TLS layer) the disparities occur.  These disparities will be worked through for the different (D)TLS libraries that support PKI so that what occurs is consistent (with potentially modified) libcoap documentation.

There are various flags that can be set or unset when setting up a PKI session and I have added some additional clarity
````
  uint8_t verify_peer_cert;        /**< 1 if peer cert is to be verified */
  uint8_t require_peer_cert;       /**< 1 if peer cert is required */
  uint8_t allow_self_signed;       /**< 1 if self signed certs are allowed */
  uint8_t allow_expired_certs;     /**< 1 if expired certs are allowed */
  uint8_t cert_chain_validation;   /**< 1 if to check cert_chain_verify_depth */
  uint8_t cert_chain_verify_depth; /**< recommended depth is 3 */
  uint8_t check_cert_revocation;   /**< 1 if revocation checks wanted */
  uint8_t allow_no_crl;            /**< 1 ignore if CRL not there */
  uint8_t allow_expired_crl;       /**< 1 if expired crl is allowed */
````
`verify_peer_cert` set causes the logic to check that both certificates are signed by a common CA, if not this is a failure, so un-setting this in Peter's case is what is required.  If a peer certificate is not received, then this CA test does not take place.

`require_peer_cert `set instructs the server to request the client to send its certificate.  If the client fails to send a certificate, then this is a failure.  Not setting this on a client, but `verify_peer_cert `is set is where the ambiguities come in and needs to be fixed.

`allow_self_signed` needs to be set if there is any certificate that is set in the CA chain that is self-signed (which also includes the CA) and `verify_peer_cert `is set.

`cert_chain_verify_depth `is the certificate chain length back to the common CA (count includes both the peer cert and its CA) including any interim CA.

obgm:
As PR #561 has been merged, I think this can be closed off, correct?

mrdeep1:
Yes, this can be closed

PR #565, opened 2020-11-12
Allow network interface to be specified when joining multicast group

szekelyisz:
This change renames `coap_join_mcast_group()` to `coap_join_mcast_group_netif()`
and adds an extra `unsigned int intf_index` parameter to specify the network
interface, which is passed to `setsockopt()`. If it is zero, then it falls back
to resolving `::` as before.

compatibility.

### DIFF
--- a/include/coap2/net.h
+++ b/include/coap2/net.h
@@ -721,11 +721,16 @@ unsigned int coap_calc_timeout(coap_session_t *session, unsigned char r);
  *
  * @param ctx   The current context
  * @param groupname The name of the group that is to be joined for listening
+ * @param intf_index Index of network interface to join the group on
  *
  * @return       0 on success, -1 on error
  */
 int
-coap_join_mcast_group(coap_context_t *ctx, const char *groupname);
+coap_join_mcast_group_intf(coap_context_t *ctx, const char *groupname,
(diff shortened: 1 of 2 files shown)


mrdeep1:
You need to run `make update-map-file`  and include `libcoap-2.map` and` libcoap-2.sym` in the commit to include any new externally visible functions.

szekelyisz:
So I did update the map file on a new branch but now the diff also includes the addition of `coap_tick_t` in the map which has nothing to do with my changes, but the commits before. Maybe somebody forgot the do the same on the commit my branch is based on? I need a bit of advice how to clean this up.

mrdeep1:
`coap_tick_t` addition is weird - all I can think of is that your version of `ctags` is not playing ball.  What system are you developing on?

I would have expected a single change in `libcoap-2.{map|sym}` which is `coap_join_mcast_group` replaced by `coap_join_mcast_group_netif` as that is not the externally visible function name.

szekelyisz:
Yeah that was my expectation, too. You're right, I had [universal-ctags](https://ctags.io/) installed. Retrying it with good old exuberant-ctags removed `coap_tick_t` from the map files.

PR #570, opened 2020-11-14
Allow network interface to be specified when joining multicast group

szekelyisz:
This change renames `coap_join_mcast_group()` to `coap_join_mcast_group_intf()`
and adds an extra `unsigned int intf_index` parameter to specify the network
interface, which is passed to `setsockopt()`. If it is zero, then it falls back
to resolving `::` as before.

compatibility.


obgm:
Closed as superseded by PR #587 which went into merged PR #617.

PR #571, opened 2020-11-14
refactor: replace magic numbers with defines

Delsian:
CoAP response codes in project represented by non-human-readable magic
numbers. Magic numbers in code where replaced by human-readable enums.
Didn't touch macro COAP_RESPONSE_CODE to minimize changes.

Signed-off-by: Eug Krashtan <eug.krashtan@gmail.com>


mrdeep1:
Thank you for looking into this.

My immediate gut feeling is that instead of doing, say,
````
-    code = COAP_RESPONSE_CODE(503);
+    code = COAP_RESPONSE_CODE(COAP_RESPONSE_CODE_SERVICE_UNAVAILABLE);
`````
you just did (which does not change the number of changes, but makes it much more readable)
````
-    code = COAP_RESPONSE_CODE(503);
+    code = COAP_RESPONSE_CODE_SERVICE_UNAVAILABLE;
````
and had
````
+ enum coap_response_code {
+	COAP_RESPONSE_CODE_OK = COAP_RESPONSE_CODE(200),
       ....
````
This also means that you are less likely to exceed the line length of 80 characters as indicated in the coding style.

**Comment by Delsian** [comment]
@mrdeep1 , totally agree, but in net.c we have some pieces of code with setting variables before calling COAP_RESPONSE_CODE() macro. So I prefer to actually implement define in this way.

**Comment by Delsian** [comment]
@mrdeep1 but yes, you're absolutely right about line length. Well, I kept net.c unchanged and replace macro in other places.
Thanks for your comment!

**Comment by mrdeep1** [comment]
This is looking better,  Not sure why you need a new define for `coap_make_response_code()` when `COAP_RESPONSE_CODE()` effectively does the same thing.
````
enum coap_response_code {
	COAP_RESPONSE_CODE_OK = coap_make_response_code(2, 0),
      ...
````
and just go with instead
````
enum coap_response_code {
	COAP_RESPONSE_CODE_OK = COAP_RESPONSE_CODE(200),
      ...
````

**Comment by obgm** [comment]
I can live with the enums to define symbolic error codes to be used in applications. But I totally disagree with changing the error codes within the libcoap core. The reason for this is that the protocol specification and its extensions specifications most of the time talk about the numeric values and implementers (or at least one that I know) think in numeric values.

**Comment by Delsian** [comment]
@obgm , well, we kept numeric values in the header file, but whatever the purpose of using magic numbers it still considered as bad practise ( https://en.wikipedia.org/wiki/Magic_number_(programming) ):

> The term magic number also refers to the bad programming practice of using numbers directly in source code without explanation. In most cases this makes programs harder to read, understand, and maintain. Although most guides make an exception for the numbers zero and one, it is a good idea to define all other numbers in code as named constants.

With enums, it's significantly easier to read and understand response codes during the debug process.

**Comment by obgm** [comment]
In the past, I have been torturing committers to replace their magic numbers with pre-processor defines or enums. This can be treated as an indicator that I am aware of magic numbers being bad practice. However, the point that I have made in my previous comment was that I do not think that numeric return codes within the library core are magic. End of discussion.

**Comment by Delsian** [comment]
Ok. Left core part untouched, only add this feature to end-users.

**Comment by obgm** [comment]
Thank you, this looks good now. I suggest adding 4.09, 4.22, 4,29 to the coap_response_code enum as these are the only missing from the current  [IANA list of CoAP response codes](https://www.iana.org/assignments/core-parameters/core-parameters.xhtml#response-codes),

File: include/coap2/pdu.h [hunk 1]
```diff
@@ -346,6 +346,44 @@ typedef uint8_t coap_proto_t;
...
+	COAP_RESPONSE_CODE_CONTINUE = COAP_RESPONSE_CODE(231),
+	COAP_RESPONSE_CODE_BAD_REQUEST = COAP_RESPONSE_CODE(400),
+	COAP_RESPONSE_CODE_UNAUTHORIZED = COAP_RESPONSE_CODE(401),
+	COAP_RESPONSE_CODE_BAD_OPTION = COAP_RESPONSE_CODE(402),
+	COAP_RESPONSE_CODE_FORBIDDEN = COAP_RESPONSE_CODE(403),
+	COAP_RESPONSE_CODE_NOT_FOUND = COAP_RESPONSE_CODE(404),
+	COAP_RESPONSE_CODE_NOT_ALLOWED = COAP_RESPONSE_CODE(405),
+	COAP_RESPONSE_CODE_NOT_ACCEPTABLE = COAP_RESPONSE_CODE(406),
+	COAP_RESPONSE_CODE_INCOMPLETE = COAP_RESPONSE_CODE(408),
+  COAP_RESPONSE_CODE_CONFLICT = COAP_RESPONSE_CODE(409),
```
obgm, line 370: There seem to be a few whitespace issues in this list.

obgm (commented):
These changes would have to be reflected in the manuals as well

```
$ grep -n COAP_RESPONSE_CODE man/*.txt.in
man/coap_cache.txt.in
man/coap_handler.txt.in
man/coap_observe.txt.in
man/coap_pdu_setup.txt.in
man/coap_resource.txt.in
```

Delsian:
@obgm 
Ok, agree

include/coap2/pdu.h
```diff
@@ -346,6 +346,44 @@ typedef uint8_t coap_proto_t;
...
 #define COAP_PROTO_TLS          4
 
+/**
+ * @brief Set of response codes available for a response packet.
+ *
+ * To be used when creating a response.
+ */
+enum coap_response_code {
+	COAP_RESPONSE_CODE_OK = COAP_RESPONSE_CODE(200),
+	COAP_RESPONSE_CODE_CREATED = COAP_RESPONSE_CODE(201),
```
mrdeep1, line 356: @Delsian Following @obgm's comment I would expect that all of the COAP_RESPONSE_ to be indented by 2 characters, not 6 characters for this enum.

PR #587, opened 2020-12-20
Allow network interface name to be specified when joining multicast group

### LINKED ISSUE #566: Is multicast server supported only on IPv6?
It looks like `coap_join_mcast_group()` is IPv6 only. Is that intentional? Should it just work on v4 as well? Any gotchas implementing it for v4?

szekelyisz:
This change renames `coap_join_mcast_group` to `coap_join_mcast_group_intf`
and adds an extra `const char *ifname` parameter to specify the interface
name.

compatibility.

`coap_is_mcast` has been modified to recognize IPv6-mapped IPv4 multicast
addresses. Without this `coap_network_send` treats these addresses as
unicast which makes `semdmsg` fail with EINVAL.

fixes #570 
fixes #566

### DIFF
--- a/include/coap2/net.h
+++ b/include/coap2/net.h
@@ -716,13 +716,18 @@ unsigned int coap_calc_timeout(coap_session_t *session, unsigned char r);
 /**
  * Function interface for joining a multicast group for listening
  *
- * @param ctx   The current context
+ * @param ctx       The current context
  * @param groupname The name of the group that is to be joined for listening
+ * @param ifname    Network interface to join the group on
  *
  * @return       0 on success, -1 on error
  */
 int
-coap_join_mcast_group(coap_context_t *ctx, const char *groupname);
+coap_join_mcast_group_intf(coap_context_t *ctx, const char *groupname,
+                           const char *ifname);
+
+#define coap_join_mcast_group(ctx, groupname) \
+	    (coap_join_mcast_group_intf(ctx, groupname, NULL))
 
 /**
  * @defgroup app_io Application I/O Handling
--- a/src/net.c
+++ b/src/net.c
@@ -27,11 +27,17 @@
 #ifdef HAVE_SYS_SOCKET_H
 #include <sys/socket.h>
 #endif
+#ifdef HAVE_SYS_IOCTL_H
+#include <sys/ioctl.h>
+#endif
 #ifdef HAVE_NETINET_IN_H
 #include <netinet/in.h>
 #endif
 #ifdef HAVE_ARPA_INET_H
 #include <arpa/inet.h>
+#endif
+#ifdef HAVE_NET_IF_H
+#include <net/if.h>
 #endif
 #ifdef COAP_EPOLL_SUPPORT
 #include <sys/epoll.h>
@@ -2939,37 +2945,70 @@ void coap_cleanup(void) {
 
 #if ! defined WITH_CONTIKI && ! defined WITH_LWIP && ! defined RIOT_VERSION
 int
-coap_join_mcast_group(coap_context_t *ctx, const char *group_name) {
-  struct ipv6_mreq mreq;
+coap_join_mcast_group_intf(coap_context_t *ctx, const char *group_name,
+  const char *ifname) {
+  struct ip_mreq mreq4;
+  struct ipv6_mreq mreq6;
   struct addrinfo   *reslocal = NULL, *resmulti = NULL, hints, *ainfo;
+  struct ifreq ifr;
   int result = -1;
   coap_endpoint_t *endpoint;
   int mgroup_setup = 0;
 
-  /* we have to resolve the link-local interface to get the interface id */
-  memset(&hints, 0, sizeof(hints));
-  hints.ai_family = AF_INET6;
-  hints.ai_socktype = SOCK_DGRAM;
+  mreq6.ipv6mr_interface = -1U;
+  mreq4.imr_interface.s_addr = INADDR_BROADCAST;
 
-  result = getaddrinfo("::", NULL, &hints, &reslocal);
-  if (result != 0) {
-    coap_log(LOG_ERR,
-             "coap_join_mcast_group: cannot resolve link-local interface: %s\n",
-             gai_strerror(result));
-    goto finish;
-  }
+  if (!ifname) {
+    /* no interface specified */
+    /* we have to resolve the link-local interface to get the interface id */
+    memset(&hints, 0, sizeof(hints));
+    hints.ai_family = AF_INET6;
+    hints.ai_socktype = SOCK_DGRAM;
 
-  /* get the first suitable interface identifier */
-  for (ainfo = reslocal; ainfo != NULL; ainfo = ainfo->ai_next) {
-    if (ainfo->ai_family == AF_INET6) {
-      mreq.ipv6mr_interface =
+    result = getaddrinfo("::", NULL, &hints, &reslocal);
+    if (result != 0) {
+      coap_log(LOG_WARNING, "coap_join_mcast_group_intf: "
+               "cannot resolve link-local interface: %s\n",
+               gai_strerror(result));
+    }
+    else {
+      /* get the first suitable interface identifier */
+      for (ainfo = reslocal; ainfo != NULL; ainfo = ainfo->ai_next) {
+        if (ainfo->ai_family == AF_INET6) {
+            mreq6.ipv6mr_interface =
                 ((struct sockaddr_in6 *)ainfo->ai_addr)->sin6_scope_id;
-      break;
+          break;
+        }
+      }
+    }
+
+    mreq4.imr_interface.s_addr = INADDR_ANY;
+  }
+  else {
+    /* interface specified */
+    /* get interface index for IPv6 and address for IPv4 */
+    strncpy(ifr.ifr_name, ifname, IFNAMSIZ - 1);
+
+    result = ioctl(ctx->endpoint->sock.fd, SIOCGIFINDEX, &ifr);
+    if (result != 0) {
+      coap_log(LOG_WARNING, "coap_join_mcast_group_intf: "
+               "cannot get interface index for '%s': %d\n", ifname, result);
+    }
+    else {
+      mreq6.ipv6mr_interface = ifr.ifr_ifindex;
+    }
+
+    result = ioctl(ctx->endpoint->sock.fd, SIOCGIFADDR, &ifr);
+    if (result != 0) {
+      coap_log(LOG_WARNING, "coap_join_mcast_group_intf: "
+               "cannot get IPv4 address for '%s': %d\n", ifname, result);
+    }
+    else {
+      mreq4.imr_interface = ((struct sockaddr_in*)&ifr.ifr_addr)->sin_addr;
     }
   }
 
   memset(&hints, 0, sizeof(hints));
-  hints.ai_family = AF_INET6;
   hints.ai_socktype = SOCK_DGRAM;
 
   /* resolve the multicast group address */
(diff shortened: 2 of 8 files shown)


mrdeep1:
Thanks for providing this fix.

I would suggest that `examples/coap-rd.c` and `examples/coap-server.c` are updated to use `coap_join_mcast_group_intf()` with a NULL value for the interface name so that future copies of this code start with the latest functionality.

As this replaces #570, I suggest that you close #570.

The build not completing would appear to be a Travis CI issue - and should be OK when you update (using squash to make it a single commit) and push this PR.

szekelyisz:
> I would suggest that `examples/coap-rd.c` and `examples/coap-server.c` are updated to use `coap_join_mcast_group_intf()` with a NULL value for the interface name so that future copies of this code start with the latest functionality.

Done.

> As this replaces #570, I suggest that you close #570.

Yes, that should happen automatically if you merge this PR, just like #566. I'll close them manually if not.

include/coap2/net.h
```diff
@@ -716,13 +716,18 @@ unsigned int coap_calc_timeout(coap_session_t *session, unsigned char r);
 /**
  * Function interface for joining a multicast group for listening
  *
- * @param ctx   The current context
+ * @param ctx       The current context
  * @param groupname The name of the group that is to be joined for listening
+ * @param ifname    Index of network interface to join the group on
```
obgm, line 721: `ifname` specifies the name of the  network interface rather than its index. I suggest: s/Index of//.

obgm (commented):
Sorry for the late review. There is just a minor comment I have before merge.

include/coap2/net.h
szekelyisz: Thanks for catching this.

src/net.c
```diff
@@ -2939,37 +2945,70 @@ void coap_cleanup(void) {
...
+    result = ioctl(ctx->endpoint->sock.fd, SIOCGIFINDEX, &ifr);
+    if (result != 0) {
+      coap_log(LOG_WARNING, "coap_join_mcast_group_intf: "
+               "cannot get interface index for '%s': %d\n", ifname, result);
+    }
+    else {
+      mreq6.ipv6mr_interface = ifr.ifr_ifindex;
+    }
+
+    result = ioctl(ctx->endpoint->sock.fd, SIOCGIFADDR, &ifr);
```
obgm, line 3001: `SIOCGIFADDR` is for `AF_INET` sockets only, therefore it will always fail.

mrdeep1:
Using this patch file correct the above and works for me  
[mcast_if.txt](https://github.com/obgm/libcoap/files/6068856/mcast_if.txt)

obgm:
@szekelyisz Does that patch work for you?

src/net.c
```diff
@@ -2939,37 +2945,70 @@ void coap_cleanup(void) {
...
+  }
+  else {
+    /* interface specified */
+    /* get interface index for IPv6 and address for IPv4 */
+    strncpy(ifr.ifr_name, ifname, IFNAMSIZ - 1);
+
+    result = ioctl(ctx->endpoint->sock.fd, SIOCGIFINDEX, &ifr);
+    if (result != 0) {
+      coap_log(LOG_WARNING, "coap_join_mcast_group_intf: "
+               "cannot get interface index for '%s': %d\n", ifname, result);
```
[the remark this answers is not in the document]
szekelyisz: That's right. Thanks for pointing out. I see they had been fixed by @mrdeep1.

src/net.c
szekelyisz: What if `coap_new_endpoint` is passed a `listen_addr` that contains an `AF_INET` address? Then the socket will also be a `AF_INET` socket.

szekelyisz:
Sorry I've been a bit busy lately. @mrdeep1 thanks for the patch. It fixes a bunch of problems. I only have one problem with it, when `getaddrinfo` returns multiple `AF_INET` results it means that many socket creations and ioctls, all ending in the same result. The original code first determines both the interface index for `mreq6` and the `AF_INET` address for `mreq4`. Then just uses whichever it needs to depending on the result being processed from `getaddrinfo`. Why does it make more sense to put this inside the loop?

szekelyisz:
Why are you guys so sure that `ctx->endpoint->sock.fd` can't be an `AF_INET` socket? Did I miss something?

obgm:
Let me put it this way: It must work for both, IPv6 and IPv4.

mrdeep1:
While obtuse, there is no reason as to why the listen address and the group multicast address cannot be a mixture of IPv4 and IPv6.  So, as obgm says, the code must be able to support both.

I take your point about the potential of sockets being opened and closed several times.    I will re-order my patch.

I have just found that `getaddrinfo("::", NULL, &hints, &reslocal);` only returns one entry, not an iteration of the interfaces and `((struct sockaddr_in6 *)ainfo->ai_addr)->sin6_scope_id` (the interface index) is left at 0, and so my linux kernel then goes on to chose the fist interface.  Unfortunately `getifaddrs()` is not POSIX compliant and cannot be used here.  It may be that we do not need to be doing `getaddrinfo("::", NULL, &hints, &reslocal);` at all and just using a ifindex of 0.

mrdeep1:
Windows build fails with the IPv4 stuff - things like `SIOCGIFINDEX `are not available.  So have put back and wrapped original code with `_WIN32 `to allow windows build to continue.

Added `-G` option to `coap-server` to test out interface binding.

Otherwise, changes are in (added in 2 functions for readability / code de-duplication) 
[mcast_if-2.patch.txt](https://github.com/obgm/libcoap/files/6095989/mcast_if-2.patch.txt)

szekelyisz:
> Let me put it this way: It must work for both, IPv6 and IPv4.

@obgm It doesn't? Sorry, I don't get what the problem is here.

My idea was that the function resolves the multicast address/DNS name to both IPv6 and IPv4 addresses, then iterates through the results and joins all resulting groups depending on the type of the original socket (whose type is unknown, so results from the other address family will fail but that's not a problem). Do you see an error in this logic or my implementation?

@mrdeep1 Thanks for adding the command line switch to the example. Your patch only uses the first result from `getaddrinfo`. The problem is that if the name has records for both address families, the call will fail if the first result's AF doesn't match the socket's. Although subsequent results' would. Iterating through all returned addresses will use all the right results (and will fail for the rest, but again, that's not an issue, just a warning is printed).

I still don't understand why it is necessary to create an `AF_INET` socket. The original socket might just be one.

> I have just found that `getaddrinfo("::", NULL, &hints, &reslocal);` only returns one entry, not an iteration of the interfaces and `((struct sockaddr_in6 *)ainfo->ai_addr)->sin6_scope_id` (the interface index) is left at 0, and so my linux kernel then goes on to chose the fist interface. Unfortunately `getifaddrs()` is not POSIX compliant and cannot be used here. It may be that we do not need to be doing `getaddrinfo("::", NULL, &hints, &reslocal);` at all and just using a ifindex of 0.

What's wrong with that? If the caller doesn't specify the interface, the kernel will pick the best one. If the caller wants to join on all interfaces, it should iterate through them. BTW `if_nameindex` is unavailable in ESP-IDF.

mrdeep1:
Looks like we need to rethink this given Windows / ESP-IDF and other potential port issues. It appears that I have made it too complicated and need to simplify it again.

So, with `coap_join_mcast_group_intf(ctx, group_name, ifname),` if `ifname` is NULL, then let the kernel choose the best one `(index = 0` (IPv6), or `addr = INADDR_ANY` (IPv4)) and support group_name being a host type entry that can be one or more of IPv4/IPv6 as well as an explicit IPv4/IPv6 definition.

I will try to do this later.

When calling `coap_new_endpoint()`, `endpoint->sock.fd` can be created as AF_INET or AF_INET6 depending on kernel support and whether the listen IP is being explicitly defined.  It is most likely to end up as AF_INET6 using examples/coap-server.c when no listen IP is defined.  Then if `group_name` is an IPv4 address (and `ifname` is defined), we run into the issues that obgm has highlighted and the code needs to be able to handle it.

mrdeep1:
Please see 
[mcast_if-3.patch.txt](https://github.com/obgm/libcoap/files/6097443/mcast_if-3.patch.txt)
for comment.  Windows does not support `struct ifreq` so no support for `ifname `at present.

szekelyisz:
> So, with `coap_join_mcast_group_intf(ctx, group_name, ifname),` if `ifname` is NULL, then let the kernel choose the best one `(index = 0` (IPv6), or `addr = INADDR_ANY` (IPv4)) and support group_name being a host type entry that can be one or more of IPv4/IPv6 as well as an explicit IPv4/IPv6 definition.

Yes, that's my idea about how this should work.

> When calling `coap_new_endpoint()`, `endpoint->sock.fd` can be created as AF_INET or AF_INET6 depending on kernel support and whether the listen IP is being explicitly defined. It is most likely to end up as AF_INET6 using examples/coap-server.c when no listen IP is defined. Then if `group_name` is an IPv4 address (and `ifname` is defined), we run into the issues that obgm has highlighted and the code needs to be able to handle it.

Hmm I just tested this with the example server and it has no problem joining a v4 multicast address with the default bind address `::` or even `::1`. Although it looks like the interface on which the group is joined is independent of the listen address. You can have a socket's listen address set to a valid interface's address and the group membership belonging to another interface. Check this out (this is without your patch yet):

```
$ ip a
[...]
2: wlp0s20f3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff
    inet 192.168.1.211/24 brd 192.168.1.255 scope global dynamic noprefixroute wlp0s20f3
       valid_lft 155sec preferred_lft 155sec
[...]
$ ./coap-server -g 239.255.0.1 -A 192.168.1.211 &
[1] 129056
$ netstat -uanp | grep coap-server
udp        0      0 192.168.1.211:5683      0.0.0.0:*                           129056/coap-server  
$ netstat -gn | grep 239.255.0.1
enx70b5e8b      1      239.255.0.1
```

The address given on the command line belongs to interface `wlp0s20f3` yet the membership is established on `enx70b5e8b`. It works the same way with `-A ::1`.

mrdeep1:
To get obgm's issue reproduced, make the coap-server ask for an interface (in my case eth0)
````
$ git diff
diff --git a/examples/coap-server.c b/examples/coap-server.c
index 47e5323..bb17c49 100644
--- a/examples/coap-server.c
+++ b/examples/coap-server.c
@@ -2746,7 +2746,7 @@ main(int argc, char **argv) {
              sizeof(cache_ignore_options)/sizeof(cache_ignore_options[0]));
   /* join multicast group if requested at command line */
   if (group)
-    coap_join_mcast_group_intf(ctx, group, NULL);
+    coap_join_mcast_group_intf(ctx, group, "eth0");

   coap_fd = coap_context_get_coap_fd(ctx);
   if (coap_fd != -1) {

$ ip a
[....]
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 10000
    link/ether xx:xx:xx:xx:xx:xx brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.189/24 brd 192.168.0.255 scope global eth0
    inet6 2001::189/96 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::20c:29ff:fe9a:7bba/64 scope link
       valid_lft forever preferred_lft forever

$ ./coap-server -g 239.255.0.1 &
[1] 11356
Mar 07 18:08:20.116 WARN coap_join_mcast_group_intf: cannot get IPv4 address for
'eth0': -1
$ netstat -uanp | grep coap-server
udp        0      0 :::5683                     :::*                                    11356/./coap-server
$ netstat -gn | grep 239.255.0.1
$ fg

$ ./coap-server -g 239.255.0.1 -A 192.168.0.189 &
[1] 11585
$ netstat -uanp | grep coap-server
udp        0      0 192.168.0.189:5683          0.0.0.0:*
       11585/./coap-server
$ netstat -gn | grep 239.255.0.1
eth0            1      239.255.0.1
$ fg


````
So `endpoint->sock.fd` works if endpoint socket is forced to be IPv4, but fails otherwise.  Note
````
Mar 07 18:08:20.116 WARN coap_join_mcast_group_intf: cannot get IPv4 address for
'eth0': -1
````
I agree that if the -A option specifies the IP address of another interface, the default multicast interface does not bind to that same interface.

mrdeep1:
Updated code so that if bound to an IPv4 interface with -A , -G not specified, the multicast group will be associated with that interface.  Unfortunately the same cannot be done with IPv6 as `sin6_scope_id `is always 0 on my system.
[mcast_if-4.patch.txt](https://github.com/obgm/libcoap/files/6098045/mcast_if-4.patch.txt)

mrdeep1:
Following a "git rebase develop" from the latest develop, the following patch is needed
[mcast_if-6.patch.txt](https://github.com/obgm/libcoap/files/6103705/mcast_if-6.patch.txt)

mrdeep1:
@szekelyisz Now that #617 (based on this code) has been merged into develop, can this PR now be closed?

obgm:
Closing off as superseded by PR #617.

PR #644, opened 2021-04-17
CMakeLists.txt: Fix interface export when DTLS is off

albanLM:
Prevents TinyDTLS includes from being passed into the interface variable when ENABLE_DTLS is OFF but USE_VENDORED_TINYDTLS is ON.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,7 +469,7 @@ target_include_directories(
          $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include/>
          $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include/>
          $<INSTALL_INTERFACE:include/>
-         $<$<BOOL:${USE_VENDORED_TINYDTLS}>:${CMAKE_BINARY_DIR}/include/tinydtls>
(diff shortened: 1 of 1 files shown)


obgm:
I wonder how this precondition should ever become valid. `HAVE_LIBTINYDTLS` to be set to 1 requires `ENABLE_DTLS` to be set. A test installation configured with -DENABLE_DTLS=OFF confirms that no tinyDTLS header files are installed.

Can you describe how to reproduce the issue that you have encountered?

albanLM:
When I first installed the library I set the -DENABLE_DTLS to OFF and still got errors due to the tinydtls library not being found. I checked the other flags, -USE_VENDORED_TINYDTLS was still on ON.
Basically what happens is: if you only put -DENABLE_DTLS=OFF, the `$<BOOL:${USE_VENDORED_TINYDTLS}>` expansion will still give you the path to an non-existant library.
That is the reason I added the `$<BOOL:${HAVE_LIBTINYDTLS}>` condition, because it is only set to 1 with -DENABLE_DTLS=ON.

obgm:
Thanks for elaborating on this. I agree that in this case the tinytdtls include path should not be appended to the target_include_directories.

PR #711, opened 2021-06-22
Fix condition for MBEDTLS_INCLUDE_DIRS

technodromega:
I had to change this in CMakeList.txt to be able to do cmake compilation with -DDTLS_BACKEND=mbedtls option


obgm:
Yes, this looks suspiciously like a copy&paste error.

PR #721, opened 2021-07-07
gnutls:  GNUTLS_CRT_RAW not defined

Delsian:
Old version of GnuTLS, used in Raspbian,  contains no definition
for GNUTLS_CRT_RAW. This patch fixes build error:
`error: ‘GNUTLS_CRT_RAW’ undeclared`

Signed-off-by: Eug Krashtan <eug.krashtan@gmail.com>

### DIFF
--- a/src/coap_gnutls.c
+++ b/src/coap_gnutls.c
@@ -68,6 +68,10 @@
                             GNUTLS_KEY_KEY_CERT_SIGN
 #endif /* GNUTLS_VERSION_NUMBER >= 0x030606 */
 
+#ifndef GNUTLS_CRT_RAW
+#define GNUTLS_CRT_RAW GNUTLS_CRT_RAWPK
+#endif /* GNUTLS_CRT_RAW */
+
 #ifdef _WIN32
 #define strcasecmp _stricmp
 #endif


src/coap_gnutls.c
```diff
@@ -68,6 +68,10 @@
                             GNUTLS_KEY_KEY_CERT_SIGN
 #endif /* GNUTLS_VERSION_NUMBER >= 0x030606 */
 
+#ifndef GNUTLS_CRT_RAW
```
obgm, line 71: I would like to tie this to the proper GnuTLS version numbers as GNUTLS_CRT_RAWPK may or may not be defined either.

obgm (commented):
In the commit message, please use the official capitalization of GnuTLS as seen in their documentation.

src/coap_gnutls.c
Delsian: @obgm There is no exact version number. According to commit 9838cc9a9b18ea5e3d94de1ba2f8bdd5013333d8 this define was removed since 3.6.4, and then restored after 3.6.12. 
So, using `ifndef` looks like simplest way to fix issue

Delsian: Version used in RPi default installation:
```
$ gnutls-cli -v
gnutls-cli 3.6.7
Copyright (C) 2000-2021 Free Software Foundation, and others, all rights reserved.
This is free software. It is licensed for use, modification and
redistribution under the terms of the GNU General Public License,
version 3 or later <http://gnu.org/licenses/gpl.html>


Please send bug reports to:  <bugs@gnutls.org>

```

obgm: Ah, I see what they did. So, yes, you are correct that the `#ifndef` is the way go.

PR #787, opened 2021-12-09
resource: change return type of coap_resource_delete

panekk:
Change return type of coap_delete_resource.

It never fails if used correctly.

### DIFF
--- a/include/coap3/resource.h
+++ b/include/coap3/resource.h
@@ -240,11 +240,8 @@ void coap_add_resource(coap_context_t *context, coap_resource_t *resource);
  *
  * @param context  The context where the resources are stored.
  * @param resource The resource to delete.
- *
- * @return         @c 1 if the resource was found (and destroyed),
- *                 @c 0 otherwise.
  */
-int coap_delete_resource(coap_context_t *context, coap_resource_t *resource);
+void coap_delete_resource(coap_context_t *context, coap_resource_t *resource);
 
 /**
  * Registers the specified @p handler as message handler for the request type
--- a/man/coap_resource.txt.in
+++ b/man/coap_resource.txt.in
@@ -39,7 +39,7 @@ _proxy_handler_, size_t _host_name_count_, const char *_host_name_list_[]);*
 *void coap_add_resource(coap_context_t *_context_,
 coap_resource_t *_resource_);*
 
-*int coap_delete_resource(coap_context_t *_context_,
+*void coap_delete_resource(coap_context_t *_context_,
 coap_resource_t *_resource_);*
 
 *void coap_resource_set_mode(coap_resource_t *_resource_, int _mode_);*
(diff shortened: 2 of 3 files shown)


obgm:
I agree with changing the return type to `void`, this makes a lot of sense.

One question is if `context == NULL || resource == NULL` should still be allowed to mimic the usual behavior of `free()`. @mrdeep1 do you have an opinion on this?

mrdeep1:
Changing the return from `int `to `void `will break the API which I am a bit loath to do.  In particular there is a (I agree outside) chance that someone is checking the return value from `coap_resopurce_delete() `and just libcoap gets updated - then random results will occur.

I do like following the `free()` model - it does mean that extra wrapping code checking variables are set before `calling coap_resource_delete()` unecessarily causes code bloat.

panekk:
Alternatively the return type may remain the same, and docs may be updated to indicate that this function never fails. Changing the return type may be considered in future, with some major update.

However, I'm not sure about checking that `context != NULL`. While it's possible that someone want to call this function directly with a result of `coap_get_resource_from_uri_path` for example, that having a NULL coap ctx is a clearly invalid state. For example `coap_delete_all_resources` doesn't check it at all.

mrdeep1:
`coap_delete_all_resources()` is for internal use only, and is not part of the public API.

With `coap_get_resource_from_uri_path()` (which is public API), I agree that there is not a check for the validity of `ctx `and would get a segmentation violation with `ctx== NULL`.

obgm:
Yes, I was more worried about `resource`, not so much about `context`. Using libcoap functions without a valid context object almost certainly is a severe error.

panekk:
Now I'm thinking, that maybe it was a bad idea at all, and instead the function should fail if the given resource doesn't belong to the given context. With the current implementation, if an invalid context is passed, the resource (with unset `is_unknown` and `is_proxy_uri`) will be freed anyway and this may not be the expected behavior. And as I see, the resource won't be removed from the list of the context it belongs to.

Which brings the question - why it's required to pass `context` to this function, if there's a `context` available inside the resource struct anyway? IMO it would be the best to break the API and change the signature to `void coap_resource_delete(coap_resource_t *resource)` (or maybe even `void coap_resource_delete(coap_resource_t **resource)` to set the passed pointer to NULL, but I think it's not a common practice in libcoap). This would also help in the case from #788

mrdeep1:
I will work on a fix to this that does not break the API (essentally the passed in ctx parameter is ignored) that also supports #788 requirements.

panekk:
I could do it as well, but as you wish :)

mrdeep1:
Interesting point about coap_free_resource v coap_delete_resource.  There is a degree of inconsistency in naming.

Thanks for your offer to do it, but I had almost finished the patch by the time I had seen your update.  Please let us know what you think about #791.

mrdeep1:
@panekk Can this be closed off now that #791 (which does not break the binary API) has been merged into the develop branch?

obgm:
If not closed off, the code should be rebased and conflicts should resolved.

PR #792, opened 2021-12-13
Corrects return value in net.c#coap_send_internal

kkrentz:
The interface description of coap_send_internal says that COAP_INVALID_MID is returned  in the case of errors. Presently, it, however, returns `(coap_mid_t)bytes_written` in a special case. Accidentally, bytes_written == -1 == COAP_INVALID_MID in this specific case, which is why no problems occur in practice due to this. As a cosmetic fix, this PR changes the return value to COAP_INVALID_MID.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1281,8 +1281,7 @@ coap_send_internal(coap_session_t *session, coap_pdu_t *pdu) {
   }
 
   if (bytes_written < 0) {
-    coap_delete_pdu(pdu);
-    return (coap_mid_t)bytes_written;
+    goto error;
   }
 
 #if !COAP_DISABLE_TCP


mrdeep1:
This change looks good to me.

obgm:
Can you please rephrase the commit message (cf. [CONTRIBUTE guidelines](https://github.com/obgm/libcoap/blob/main/CONTRIBUTE)). 

For example:

```
net.c: correct return value in coap_send_internal()
```

Moreover, the description usually should also contain the rationale for the change. (Here: Make the function return  `COAP_INVALID_MID` if `coap_send_pdu()` failed.)

This should also be stated in the PR's description to leave the reviewers to decide whether or not the fix is small. I at least wonder if `COAP_INVALID_MID` is the correct return code here. It might be more reasonable to pass through the return code of `coap_send_pdu()` (which actually will be `COAP_INVALID_MID` == -1 as far as I can see because the only other code, `COAP_PDU_DELAYED` is handled right above this change.) I am sure that you have considered this and weighed the pros and cons of changing the return code. But these thoughts should be expressed in the description of the PR and result in a short reasoning in the commit message to avoid giving developers are hard time interpreting this "small fix".

PR #802, opened 2021-12-22
Allow usage of a non-vendored TinyDTLS version when using autotools

pulsastrix:
This PR modifies the autotools scripts (mostly `configure.ac`) to allow using a tinydtls version that is different from the one provided in the submodule.

To do so, set the `--with-system-tinydtls` (as well as the `--with-tinydtls`) configure script option when building.
It is also possible to override the linker and compiler flags using the `TinyDTLS_LIBS` and `TinyDTLS_CFLAGS` environment variables, analogously to the `(OpenSSL|GnuTLS)_(LIBS|CFLAGS)` environment variable for the other TLS backends.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -318,6 +318,7 @@ AM_CONDITIONAL(BUILD_MANPAGES, [test "x$build_manpages" = "xyes"])
 gnutls_version_required=3.3.0
 openssl_version_required=1.1.0
 mbedtls_version_required=2.7.10
+tinydtls_version_required=0.8.6
 
 AC_ARG_ENABLE([dtls],
               [AS_HELP_STRING([--enable-dtls],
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -421,7 +429,26 @@ if test "x$build_dtls" = "xyes"; then
         fi
     fi
 
-    # TinyDTLS ?
+    # TinyDTLS [also does not have pkg-config file]
+    AC_CHECK_LIB(tinydtls, dtls_new_context, [have_system_tinydtls="yes"; TinyDTLS_CFLAGS=""; TinyDTLS_LIBS="-ltinydtls"], [have_system_tinydtls="no"])
```
obgm: I wonder if (parts of) this should go as a macro into a new `m4/ac_check_tinydtls.m4`.

obgm:
@tijuca Do you have an opinion if this blends well with usual package management best practices?

pulsastrix:
With eclipse/tinydtls#120 being merged, the changes could probably be simplified a bit by using pkg-config (similarly to the way the other TLS libraries are searched for). 

This would make the changes in `src/coap_tinydtls.c` made by this PR unnecessary: pkg-config would add the required include path so that tinydtls' headers could be referenced without the `tinydtls/` path component.

However, this would have the disadvantage that this would not work with older versions of tinydtls that lack the pkg-config file.

Are there any opinions on how to proceed in this matter?

mrdeep1:
I would prefer that the changes to coap_tinydtls.c remain - namely the tinydtls headers are properly prefixed.

There is a dependency build order to think about, i.e if tinydtls.pc is found, then is that to be used instead of submodule ext/tinydtls?
Or do we need a way to give priority to ext/tinydtls over tinydtls.pc?

It is straighforward to bring libcoap forward to using the latest tinydtls as a submodule (I have been doing doing this locally and no issues found so far).

obgm:
My opinion on the build order: I presume that if a user has installed a version of tinydtls, it is reasonable to assume that she wants this to be used. This fits into the existing build order in configure.ac where ext/tinydtls is the last resort if no other TLS libraries are found.

pulsastrix:
Based on your comments I have made the following changes to the PR:
- `--with-system-tinydtls` is replaced with `--with-vendored-tinydtls`. 
- The system-provided TinyDTLS can now be chosen automatically if no DTLS library was selected (the vendored version will still only be selected if `--with-tinydtls` is explicitly enabled).
- If `--with-tinydtls` is set, the configure-script will now prefer the system-provided version of tinydtls. This can be overridden by setting `--with-vendored-tinydtls=yes`.
    - It will however fall back to the vendored version if no system-provided version was found. This can also be overridden by setting `--with-vendored-tinydtls=no`
- The system-provided TinyDTLS is now detected using `pkg-config`. This way, we can avoid some problems that arise due to missing include paths (some parts of TinyDTLS itself assume that the other headers can be found without the `tinydtls/` path prefix.

This should be _mostly_ compatible with the previous behaviour of the script: As long as TinyDTLS is not installed system-wide, nothing changes. 
If TinyDTLS is installed system-wide, the behaviour will be different as long as  `--with-vendored-tinydtls` is not explicitly set. 
However, this should only affect very few people, especially considering that the installed version would have to be pretty recent to provide the necessary `pkg-config` file.

mrdeep1:
With your current code, simply changing the word `vendored `with `submodule `in `configure.ac` (so we get `--with-submodule-tinydtls` etc..) as `vendored `is ambiguous to me, but `submodule `is obvious.

The commits needs updating to reflect submodule, not system - and all the commits should be squashed into a single commit.

Otherwise the code changes work for me when the above is done.

pulsastrix:
Rebased on current state of `develop`, replaced occurrences of "vendored" with "submodule" and squashed into a single commit (excluding the submodule update as it was already cherry-picked).

Issue #765, opened 2021-10-29
mbedtls 3.0 support


boaks:
Does libcoap supports dtls bindings for mbedtls 3.0?
If not, any plans to do so?

obgm:
I have not yet looked into the details of the API changes but my expectation is that we will shortly adapt to v3.0 in addition to the v2.x bindings.

PR #822, opened 2022-02-09
`coap_mbedtls.c`: Upgrade to mbedTLS `v3.1.0`

laukik-hase:
- Added `MBEDTLS_ALLOW_PRIVATE_ACCESS` to access private struct members wherever necessary (keeping backward compatibility)
- Updated deprecated functions from hashing module (E.g. `mbedtls_sha256_starts_ret()` -> `mbedtls_sha256_starts()`)
- Added mandatory RNG parameter for some functions (`mbedtls_pk_parse_keyfile()`, `mbedtls_pk_parse_key()`)
- Removed support for parsing `SSLv2 ClientHello`
- Closes https://github.com/obgm/libcoap/issues/765

### DIFF
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -44,13 +44,23 @@
  */
 
 #include <mbedtls/version.h>
+
+/* Keep forward-compatibility with Mbed TLS 3.x */
+#if (MBEDTLS_VERSION_NUMBER < 0x03000000)
+#define MBEDTLS_2_X_COMPAT
+#else /* !(MBEDTLS_VERSION_NUMBER < 0x03000000) */
+/* Macro wrapper for struct's private members */
+#ifndef MBEDTLS_ALLOW_PRIVATE_ACCESS
+#define MBEDTLS_ALLOW_PRIVATE_ACCESS
+#endif /* MBEDTLS_ALLOW_PRIVATE_ACCESS */
+#endif /* !(MBEDTLS_VERSION_NUMBER < 0x03000000) */
+
 #include <mbedtls/platform.h>
 #include <mbedtls/net_sockets.h>
 #include <mbedtls/ssl.h>
 #include <mbedtls/entropy.h>
 #include <mbedtls/ctr_drbg.h>
 #include <mbedtls/error.h>
-#include <mbedtls/certs.h>
 #include <mbedtls/timing.h>
 #include <mbedtls/ssl_cookie.h>
 #include <mbedtls/oid.h>
@@ -147,6 +157,13 @@ typedef enum coap_enc_method_t {
   COAP_ENC_PSK,
   COAP_ENC_PKI,
 } coap_enc_method_t;
+
+#ifndef MBEDTLS_2_X_COMPAT
+static int coap_rng(void *ctx COAP_UNUSED, unsigned char *buf, size_t len)
+{
+  return coap_prng(buf, len);
+}
+#endif /* MBEDTLS_2_X_COMPAT */
 
 static int coap_dgram_read(void *ctx, unsigned char *out, size_t outl)
 {
@@ -267,7 +284,7 @@ static char*
 get_san_or_cn_from_cert(mbedtls_x509_crt *crt)
 {
   if (crt) {
-    mbedtls_asn1_named_data * cn_data;
+    const mbedtls_asn1_named_data * cn_data;
 
     if (crt->ext_types & MBEDTLS_X509_EXT_SUBJECT_ALT_NAME) {
       mbedtls_asn1_sequence *seq = &crt->subject_alt_names;
@@ -491,8 +508,14 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
         return ret;
       }
 
+#ifdef MBEDTLS_2_X_COMPAT
       ret = mbedtls_pk_parse_keyfile(private_key,
                               setup_data->pki_key.key.pem.private_key, NULL);
+#else
+      ret = mbedtls_pk_parse_keyfile(private_key,
+                              setup_data->pki_key.key.pem.private_key,
+                              NULL, coap_rng, (void *)&m_env->ctr_drbg);
+#endif /* MBEDTLS_2_X_COMPAT */
       if (ret < 0) {
         coap_log(LOG_ERR, "mbedtls_pk_parse_keyfile returned -0x%x: '%s'\n",
                  -ret, get_error_string(ret));
@@ -573,13 +596,25 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
         memcpy(buffer, setup_data->pki_key.key.pem_buf.private_key, length);
         buffer[length] = '\000';
         length++;
+#ifdef MBEDTLS_2_X_COMPAT
         ret = mbedtls_pk_parse_key(private_key, buffer, length, NULL, 0);
+#else
+        ret = mbedtls_pk_parse_key(private_key, buffer, length,
+              NULL, 0, coap_rng, (void *)&m_env->ctr_drbg);
+#endif /* MBEDTLS_2_X_COMPAT */
         mbedtls_free(buffer);
       }
       else {
+#ifdef MBEDTLS_2_X_COMPAT
         ret = mbedtls_pk_parse_key(private_key,
               setup_data->pki_key.key.pem_buf.private_key,
               setup_data->pki_key.key.pem_buf.private_key_len, NULL, 0);
+#else
+        ret = mbedtls_pk_parse_key(private_key,
+              setup_data->pki_key.key.pem_buf.private_key,
+              setup_data->pki_key.key.pem_buf.private_key_len,
+              NULL, 0, coap_rng, (void *)&m_env->ctr_drbg);
+#endif /* MBEDTLS_2_X_COMPAT */
       }
       if (ret < 0) {
         coap_log(LOG_ERR, "mbedtls_pk_parse_keyfile returned -0x%x: '%s'\n",
@@ -650,9 +685,16 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
         return ret;
       }
 
+#ifdef MBEDTLS_2_X_COMPAT
       ret = mbedtls_pk_parse_key(private_key,
               (const unsigned char *)setup_data->pki_key.key.asn1.private_key,
               setup_data->pki_key.key.asn1.private_key_len, NULL, 0);
+#else
+      ret = mbedtls_pk_parse_key(private_key,
+              (const unsigned char *)setup_data->pki_key.key.asn1.private_key,
+              setup_data->pki_key.key.asn1.private_key_len, NULL, 0, coap_rng,
+              (void *)&m_env->ctr_drbg);
+#endif /* MBEDTLS_2_X_COMPAT */
       if (ret < 0) {
         coap_log(LOG_ERR, "mbedtls_pk_parse_keyfile returned -0x%x: '%s'\n",
                  -ret, get_error_string(ret));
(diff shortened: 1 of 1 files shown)


mrdeep1:
@laukik-hase Many thanks for doing this initial work.  However, compliling your code against older versions of Mbed TLS throws up errors  where the new functionality is not supported, or have a different parameter counts.  You therefore need to wrap  some of your code differences with
````
#if MBEDTLS_VERSION_NUMBER >= 0x03000000
...
#else /* MBEDTLS_VERSION_NUMBER < 0x03000000 */
...
#endif /* MBEDTLS_VERSION_NUMBER < 0x03000000 */
````
 (Adjusting version number as appropriate to when the feature became available).
Thought also needs to be given at to how MBEDTLS_PRIVATE() is handled for older Mbed TLS versions.

I get the following build errors
````
src/coap_mbedtls.c:154: warning: no previous prototype for â€˜coap_rngâ€™
src/coap_mbedtls.c:158:6: error: #elif with no expression
src/coap_mbedtls.c: In function â€˜coap_rngâ€™:
src/coap_mbedtls.c:154: warning: unused parameter â€˜ctxâ€™
src/coap_mbedtls.c:154: warning: unused parameter â€˜bufâ€™
src/coap_mbedtls.c:154: warning: unused parameter â€˜lenâ€™
src/coap_mbedtls.c: In function â€˜get_san_or_cn_from_certâ€™:
src/coap_mbedtls.c:284: error: â€˜mbedtls_x509_crtâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c:284: error: â€˜ext_typesâ€™ undeclared (first use in this function)
src/coap_mbedtls.c:284: error: (Each undeclared identifier is reported only once
src/coap_mbedtls.c:284: error: for each function it appears in.)
src/coap_mbedtls.c: In function â€˜setup_pki_credentialsâ€™:
src/coap_mbedtls.c:508: error: too many arguments to function â€˜mbedtls_pk_parse_keyfileâ€™
src/coap_mbedtls.c:590: error: too many arguments to function â€˜mbedtls_pk_parse_keyâ€™
src/coap_mbedtls.c:597: error: too many arguments to function â€˜mbedtls_pk_parse_keyâ€™
src/coap_mbedtls.c:671: error: too many arguments to function â€˜mbedtls_pk_parse_keyâ€™
src/coap_mbedtls.c: In function â€˜set_ciphersuitesâ€™:
src/coap_mbedtls.c:1003: error: â€˜mbedtls_ssl_ciphersuite_tâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c:1003: error: â€˜max_minor_verâ€™ undeclared (first use in this function)
src/coap_mbedtls.c:1037: error: â€˜mbedtls_ssl_ciphersuite_tâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c: In function â€˜do_mbedtls_handshakeâ€™:
src/coap_mbedtls.c:1242: error: â€˜mbedtls_ssl_contextâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c:1242: error: â€˜in_msgâ€™ undeclared (first use in this function)
src/coap_mbedtls.c:1243: error: â€˜mbedtls_ssl_contextâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c:1243: error: â€˜mbedtls_ssl_contextâ€™ has no member named â€˜MBEDTLS_PRIVATEâ€™
src/coap_mbedtls.c: In function â€˜coap_digest_updateâ€™:
src/coap_mbedtls.c:2403: error: void value not ignored as it ought to be
src/coap_mbedtls.c: In function â€˜coap_digest_finalâ€™:
src/coap_mbedtls.c:2411: error: void value not ignored as it ought to be
````
which all need to be corrected for older builds.

**Comment by mrdeep1** [comment]
@obgm Are you able to approve this workflow?

**Comment by obgm** [comment]
Yes, of course. Here we go...

**Comment by mrdeep1** [comment]
@laukik-hase That looks a lot better, thanks.  Just some minor comments.

**Comment by obgm** [comment]
Yes, great work. Hang on while I am going to update my Mbed TLS version...

**Comment by mrdeep1** [comment]
It is relatively easy to get a local copy of MbedTLS that is 3.1.0 to do a trial build against, but the github workflows would only appear to support a hosted runner that has MBed TLS < 3.0.0.

**Comment by mrdeep1** [comment]
#include <mbedtls/certs.h>
is also not available for MbedTLS 3.1.0

File: src/coap_mbedtls.c [hunk 1]
```diff
@@ -148,6 +161,17 @@ typedef enum coap_enc_method_t {
   COAP_ENC_PKI,
 } coap_enc_method_t;
 
+#ifndef MBEDTLS_2_x
+int coap_rng(void *ctx, unsigned char *buf, size_t len)
```
obgm, line 165: I would prefer not to define this function here and rely on the existing function `coap_prng()`. First of all, this function here is internal, so it should be made static (this would avoid also the [build error](https://github.com/obgm/libcoap/runs/5138665747?check_suite_focus=true#step:7:54). But then, there will be two independent sources of random data in the library: this function and the "official" function `coap_prng()`. If an application decides to set its own different entropy source using `coap_set_prng()` and happens to use Mbed TLS, it ends up with "wrong" random data for the Mbed TLS specific part.

I see why you might want to have this function. In this case, I suggest splitting it off from this PR and provide (configurable) changes against [`coap_prng_default()`](https://github.com/obgm/libcoap/blob/f9c3e13a1802e9d1e9c4ea87763133192990327a/src/coap_prng.c#L52).

laukik-hase: Hello, @obgm!
Thanks for pointing out these mistakes and suggestions. Does the following patch work?

```patch
diff --git a/src/coap_mbedtls.c b/src/coap_mbedtls.c
index d3dfa72..276023f 100644
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -148,6 +148,13 @@ typedef enum coap_enc_method_t {
   COAP_ENC_PKI,
 } coap_enc_method_t;

+#ifndef MBEDTLS_2_x
+static int coap_rng(void *ctx, unsigned char *buf, size_t len)
+{
+  return coap_prng(buf, len);
+}
+#endif
+
 static int coap_dgram_read(void *ctx, unsigned char *out, size_t outl)
 {
   ssize_t ret = 0;
diff --git a/src/coap_prng.c b/src/coap_prng.c
index e6ba61d..c3317f2 100644
--- a/src/coap_prng.c
+++ b/src/coap_prng.c
@@ -22,6 +22,10 @@
 #include <stdlib.h>
 #endif /* !HAVE_GETRANDOM */

+#if defined(MBEDTLS_ENTROPY_HARDWARE_ALT)
+#include <mbedtls/library/entropy_poll.h>
+#endif /* MBEDTLS_ENTROPY_HARDWARE_ALT */
+
 #if defined(_WIN32)

 errno_t __cdecl rand_s( _Out_ unsigned int* _RandomValue );
@@ -53,6 +57,9 @@ coap_prng_default(void *buf, size_t len) {
 #ifdef HAVE_GETRANDOM
   return getrandom(buf, len, 0);
 #else /* !HAVE_GETRANDOM */
+#if defined(MBEDTLS_ENTROPY_HARDWARE_ALT)
+  return mbedtls_hardware_poll(ctx, buf, len, NULL);
+#else
 #if defined(_WIN32)
   return coap_prng_impl(buf,len);
 #else /* !_WIN32 */

```

laukik-hase: Hello, @obgm!

Thank you for pointing out the mistakes and suggestions. Does the following patch work?
```patch
diff --git a/src/coap_mbedtls.c b/src/coap_mbedtls.c
index d3dfa72..276023f 100644
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -148,6 +148,13 @@ typedef enum coap_enc_method_t {
   COAP_ENC_PKI,
 } coap_enc_method_t;

+#ifndef MBEDTLS_2_x
+static int coap_rng(void *ctx, unsigned char *buf, size_t len)
+{
+  return coap_prng(buf, len);
+}
+#endif
+
 static int coap_dgram_read(void *ctx, unsigned char *out, size_t outl)
 {
   ssize_t ret = 0;
diff --git a/src/coap_prng.c b/src/coap_prng.c
index e6ba61d..c3317f2 100644
--- a/src/coap_prng.c
+++ b/src/coap_prng.c
@@ -22,6 +22,10 @@
 #include <stdlib.h>
 #endif /* !HAVE_GETRANDOM */

+#if defined(MBEDTLS_ENTROPY_HARDWARE_ALT)
+#include <mbedtls/library/entropy_poll.h>
+#endif /* MBEDTLS_ENTROPY_HARDWARE_ALT */
+
 #if defined(_WIN32)

 errno_t __cdecl rand_s( _Out_ unsigned int* _RandomValue );
@@ -53,6 +57,9 @@ coap_prng_default(void *buf, size_t len) {
 #ifdef HAVE_GETRANDOM
   return getrandom(buf, len, 0);
 #else /* !HAVE_GETRANDOM */
+#if defined(MBEDTLS_ENTROPY_HARDWARE_ALT)
+  return mbedtls_hardware_poll(ctx, buf, len, NULL);
+#else
 #if defined(_WIN32)
   return coap_prng_impl(buf,len);
 #else /* !_WIN32 */

```

laukik-hase: Hello, @obgm!

Concerning your comment below, any suggestions on how to split this part off?

PS: I tested this change on the ESP-IDF [master](https://github.com/espressif/esp-idf/commit/5810ed1d043cbc54c9a7616db1e84d99f71f3ccb) with mbedtls-3.1.0 and it worked just fine.

mrdeep1:
@laukik-hase You may find it easier to work through the suggestions to the code using the "Files Changed" Tab.

There are some code changes needed for 3.0.0 (`MBEDTLS_PRIVATE`) and others for 3.1.1 (`mbedtls_pk_parse_key*()` functions) to support the general upgrade, and then separately a fair number of changes for providing an alternative hardware prng which should be in a separate PR.

mrdeep1:
@laukik-hase Are you able to make the suggested changes?

laukik-hase:
Hello, @mrdeep1!

I have updated the PR, could you please take a look? I will also create a separate PR for the RNG related changes.

__Note__: These changes build successfully for mbedTLS `v2.28.0` and `v3.1.0`, but fail to build for `v3.0.0`. This is because some of the private struct fields were made public in the update.

mrdeep1:
It would be good if you opened up the "Files changed" tab and work through the different comments, updating your code as appropriate.

In terms of V3.0.0, my suggestion is that you define a MBEDTLS_PRIV_30 macro along the lines of MDEVTLS_PRIVATE (for >= 0x300000 and < 0x301000) and use it for all the definitions that were made leading private_ for v3.0.0 and then have the leading private_ removed for v3.1.0.  This would then mean that v2.x, v3.0 and v3.1 would be supported.

mrdeep1:
This looks a lot better and a neat version independent way of handling the private members..  However, there is still the following error for a 3.0.0 or 3.1.0 build
````
  CC       src/libcoap_3-mbedtls_la-coap_mbedtls.lo
src/coap_mbedtls.c: In function ‘coap_rng’:
src/coap_mbedtls.c:162:27: warning: unused parameter ‘ctx’ [-Wunused-parameter]
  162 | static int coap_rng(void *ctx, unsigned char *buf, size_t len)
      |                     ~~~~~~^~~
````
which needs to be corrected using `COAP_UNUSED` or` (void)`.

mrdeep1:
Two things

Can you please update the commit message to reflect what is now happening (i.e. MBEDTLS_PRIVATE is no longer relevant as you are conditionally adding in MBEDTLS_ALLOW_PRIVATE_ACCESS, `Added macros for maintaining backward compatibility` is no longer true etc.).

I would have added in the following code for 2.x for a reason when doing Mbed TLS testing, and so it should remain, but wrapped with #ifdef MBEDTLS_2_X_COMPAT
````
   case MBEDTLS_ERR_SSL_NO_CLIENT_CERTIFICATE:
     alert = MBEDTLS_SSL_ALERT_MSG_NO_CERT;
     goto fail_alert;
-  case MBEDTLS_ERR_SSL_BAD_HS_CLIENT_HELLO:
-    alert = MBEDTLS_SSL_ALERT_MSG_HANDSHAKE_FAILURE;
-    goto fail_alert;
   case MBEDTLS_ERR_X509_CERT_VERIFY_FAILED:
     goto fail;
   case MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE:
````

Thanks

mrdeep1:
Thanks.  This PR is now good from my perspective.

obgm:
Thanks @laukik-hase and @mrdeep1 for sorting this out!

PR #828, opened 2022-02-17
`coap_prng.c`: Added alternate RNG implementation

laukik-hase:
- For targets having their hardware RNG implementation using `mbedtls_hardware_poll()`
- This change was made because, for `mbedtls-3.x`, passing an RNG function to all functions that accept an `f_rng` parameter is mandatory
- Related to #822


mrdeep1:
That looks much better - t hanks.

PR #832, opened 2022-02-21
block.c: Fix possible null-pointer dereference

obgm:
Commit 9cb2f62a3 [1] has introduced library-internal handling of
Block1 and Block2 requests. To add the data to the response body the
function coap_add_data_large_internal() is used.

This change calls the function independent of a Block2 request unless
request is NULL. The response code is set to 500 if
coap_add_data_large_internal() failed.

Fixes Coverity-Id 1514379

[1] https://github.com/obgm/libcoap/commit/9cb2f62a3eef2e0aeb682615b785d4734b473bd4

### DIFF
--- a/src/block.c
+++ b/src/block.c
@@ -698,25 +698,15 @@ coap_add_data_large_response(coap_resource_t *resource,
     default:                        /* everything is good */
         ;
     }
-
-    if (!coap_add_data_large_internal(session, response, resource, query,
-                                      maxage, etag, length, data,
-                                      release_func, app_ptr,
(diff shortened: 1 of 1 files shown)


obgm:
The previous version has set the response code to 500 in case add_data_internal has failed when a block was requested, and to 400 in case add_data_internal has failed when Block2 was not requested. I have the feeling that both cases can be handled with 500 but I am not sure about that.

mrdeep1:
Yes, both cases should have returning 500.  PR looks good and ready to go

PR #857, opened 2022-06-07
net.c: fix null pointer exception

kkrentz:
In `net.c`, there is a place where `coap_pdu_init` is called without checking its result. This is likely to cause segmentation faults when the memory gets scarce. My fix inserts a check.


mrdeep1:
As it happens, the `coap_add_token()` a few lines on handles this condition, but setting `response->type = COAP_MESSAGE_CON;` on the following line after the patch is the cause of the segmentation fault..

Fix is good for me.

PR #934, opened 2022-09-28
Small clean-up of preprocessor directives

kkrentz:
This PR extracts two more general fixes from #760.


src/resource.c
```diff
@@ -548,10 +548,9 @@ coap_free_resource(coap_resource_t *resource) {
 
 #ifdef WITH_LWIP
   memp_free(MEMP_COAP_RESOURCE, resource);
-#endif
-#ifndef WITH_LWIP
+#else
   coap_free_type(COAP_RESOURCE, resource);
-#endif /* WITH_CONTIKI */
+#endif
```
mrdeep1, line 553: This actually should be (as` coap_free_type(type, p)` is mapped to `memp_free(MEMP_ ## type, p)` for LwIP)
````
@@ -526,12 +526,7 @@ coap_free_resource(coap_resource_t *resource) {
     coap_free(resource->proxy_name_list);
   }
 
-#ifdef WITH_LWIP
-  memp_free(MEMP_COAP_RESOURCE, resource);
-#endif
-#ifndef WITH_LWIP
   coap_free_type(COAP_RESOURCE, resource);
-#endif /* WITH_CONTIKI */
 }
 
 void
````

File: src/coap_io.c [hunk 2]
```diff
@@ -187,12 +187,10 @@ coap_socket_bind_udp(coap_socket_t *sock,
          "coap_socket_bind_udp: ioctl FIONBIO: %s\n", coap_socket_strerror());
   }
 
-#ifndef RIOT_VERSION
```
mrdeep1, line 190: Why is this not needed (I don't know enough about RIOT)?

kkrentz: This snippet is not compiled if RIOT is off anyway due to the `#ifndef RIOT_VERSION` on Line 180.

mrdeep1:
If these changes could be squashed into a single commit, that would be great.

src/resource.c
mrdeep1: You have additionally found and fixed the` coap_free_type(COAP_RESOURCEATTR, attr);` - thanks, but not the `coap_free_type(COAP_RESOURCE, resource);` as suggested above.

PR #951, opened 2022-10-18
uri: include coap_option.h

ncmiller:
Required for usage of coap_optlist_t.

### DIFF
--- a/include/coap3/uri.h
+++ b/include/coap3/uri.h
@@ -20,6 +20,7 @@
 #include <stdint.h>
 
 #include "str.h"
+#include "coap_option.h"
 
 /**
  * The scheme specifiers. Secure schemes have an odd numeric value,


mrdeep1:
@ncmiller I have no issues with this change, but am intrigued by why it is needed.

I agree that `coap_uri_into_options()` has been added recently to `uri.h`, but I would be expecting any application to be including `"coap3/coap.h"` where including `uri.h` is defined after `coap_option.h` and I cannot find another include file that includes `uri.h`.  Also the github workflow build tests did not highlight any issues.

mrdeep1:
I wonder if this is the trigger issue
````
diff --git a/include/coap3/pdu.h b/include/coap3/pdu.h
index 29331b3..f2ee665 100644
--- a/include/coap3/pdu.h
+++ b/include/coap3/pdu.h
@@ -18,8 +18,8 @@
 #ifndef COAP_PDU_H_
 #define COAP_PDU_H_
 
-#include "uri.h"
 #include "coap_option.h"
+#include "uri.h"
 
 #ifdef WITH_LWIP
 #include <lwip/pbuf.h>
````

**Comment by ncmiller** [comment]
Okay, I see the issue now. 

I'm maintaining a port of libcoap, and I'm overriding `coap3/coap.h` with my own version, which has an order of includes that doesn't match coap.h.in:

```
#include "coap3/coap_forward_decls.h"
#include "coap3/coap_address.h"
#include "coap3/coap_async.h"
#include "coap3/block.h"
#include "coap3/coap_cache.h"
#include "coap3/coap_dtls.h"
#include "coap3/coap_event.h"
#include "coap3/coap_io.h"
#include "coap3/coap_time.h"
#include "coap3/coap_debug.h"
#include "coap3/encode.h"
#include "coap3/mem.h"
#include "coap3/net.h"
#include "coap3/coap_option.h"
#include "coap3/pdu.h"
#include "coap3/coap_prng.h"
#include "coap3/resource.h"
#include "coap3/str.h"
#include "coap3/coap_subscribe.h"
#include "coap3/uri.h"
```

If I reorder them to match the order from coap.h.in, everything compiles fine for me:

```
#include "coap3/coap_forward_decls.h"
#include "coap3/block.h"
#include "coap3/coap_address.h"
#include "coap3/coap_async.h"
#include "coap3/coap_cache.h"
#include "coap3/coap_debug.h"
#include "coap3/coap_dtls.h"
#include "coap3/coap_event.h"
#include "coap3/coap_io.h"
#include "coap3/coap_prng.h"
#include "coap3/coap_option.h"
#include "coap3/coap_subscribe.h"
#include "coap3/coap_time.h"
#include "coap3/encode.h"
#include "coap3/mem.h"
#include "coap3/net.h"
#include "coap3/pdu.h"
#include "coap3/resource.h"
#include "coap3/str.h"
#include "coap3/uri.h"
```

So I suppose I could just fix this in my port instead of adding an include to uri.h. On the other hand, it's a fairly innocuous change that could remove a tiny bit of friction for people that are in a similar situation to me (dealing with order of inclusion issues in ports).

ncmiller:
> It would be good to know if there are significant differences / new functionality in your port which may have benefit for libcoap in general.

My port is is nearly identical to the Espressif port: https://github.com/espressif/idf-extra-components/tree/master/coap. 

The only reason I'm not using the Espressif port is because it's not so straightforward to pull into my project. I would either have to include the `espressif/idf-extra-components` repo as a submodule (which has a bunch of other stuff I don't need), or use the new Espressif component registry to pull in just the `coap` component, but the component registry is still very new and not part of a mainline ESP-IDF release (AFAIK). So I will continue to maintain a port until ESP-IDF releases with support for the component registry (probably 5.x.x release), and then switch over to the Espressif port.

From what I can tell, there isn't much in terms of new functionality in the port. It's mostly just header and config overrides to get things to build cleanly and work in ESP-IDF.

mrdeep1:
OK - ESP_IDF port requirements make sense.  At some point I will be submitting an update for `espressif/idf-extra-components` when libcoap gets to another release point.

mrdeep1:
See #953 for a more comprehensive fix which supersedes this PR.  Thanks for raising it.

PR #952, opened 2022-10-19
coap_event: Add COAP_EVENT_MSG_RETRANSMITTED

ncmiller:
Apart from debug log messages, there is no way for a user to be notified when a message retransmission happens.

This event can be useful for diagnosing issues related to lossy networks or unresponsive CoAP servers (e.g. user can track total retransmissions as an indicator for general link quality).

Assigned event ID to 0x5002 to avoid creating a new 0x6000 event group, and because it's similar to COAP_EVENT_BAD_PACKET (0x5001).


mrdeep1:
This looks good to me.

PR #972, opened 2022-11-20
github workflow: add CIFuzz Github action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where libcoap already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
```
mrdeep1, line 2: To prevent this workflow running twice every-time the code is updated (following main.yml), I would prefer this to be
````
on:
  push:
    branches:
      - main
      - develop
      - release-*
      - gh-workflows
  pull_request:
    branches:
      - main
      - develop
````

mrdeep1:
Unless I am mistaken, cifuzz.yml only loads in the latest develop branch, not the PR branch.

PR #984, opened 2022-12-01
Adjust Modlines in Manpage templates

tijuca:
Small adjustment so vim isn't complaining about wrong modlines any more.


mrdeep1:
Thanks for this.  I will resubmit the workflow after giving ubuntu time to recover.

PR #985, opened 2022-12-01
Spelling fixes

tijuca:
The Lintian tool in Debian did detect some misspelled words.


mrdeep1:
Thanks for this.  I will resubmit the workflow after giving ubuntu time to recover.

PR #1006, opened 2022-12-23
resource.c: Add COAP_RESOURCE_MAX_SUBSCRIBER to limit subscribers

rretanubun:
resource.c: Add COAP_RESOURCE_MAX_SUBSCRIBER to limit subscribers
    
    If COAP_RESOURCE_MAX_SUBSCRIBER is > 0
    coap_add_observer() now checks and returns NULL
    if COAP_RESOURCE_MAX_SUBSCRIBER is exceeded.
    
    net.c then detects this and respond.
    
    Per RFC7641 4.1. Request:
    
      A server that is unable or unwilling to add a new entry to the list
      of observers of a resource MAY silently ignore the registration
      request and process the GET request as usual.  The resulting response
      MUST NOT include an Observe Option, the absence of which signals to
      the client that it will not be notified of changes to the resource
      and, e.g., needs to poll the resource for its state instead.
    
    Useful for constrained systems that wants to support
    observers but need to be mindful of memory usage.

### DIFF
--- a/src/resource.c
+++ b/src/resource.c
@@ -763,6 +763,15 @@ static const uint16_t cache_ignore_options[] = { COAP_OPTION_ETAG };
     return s;
   }
 
+  /* Check if there is already maximum number of subscribers present */
+#if (COAP_RESOURCE_MAX_SUBSCRIBER > 0)
+  uint32_t subscriber_count = 0;
+  LL_COUNT(resource->subscribers, s, subscriber_count);
+  if (subscriber_count >= COAP_RESOURCE_MAX_SUBSCRIBER) {
+    return NULL; /* Signal error */
+  }
+#endif /* COAP_RESOURCE_MAX_SUBSCRIBER */
+
   /* Create a new subscription */
   s = coap_malloc_type(COAP_SUBSCRIPTION, sizeof(coap_subscription_t));
 
--- a/include/coap3/coap_resource_internal.h
+++ b/include/coap3/coap_resource_internal.h
@@ -29,6 +29,14 @@
  */
 
 /**
+ * Limits the number of subscribers for each resource that this server support.
+ * Zero means there is no maximum.
+ */
+#ifndef COAP_RESOURCE_MAX_SUBSCRIBER
+#define COAP_RESOURCE_MAX_SUBSCRIBER 0
+#endif /* COAP_RESOURCE_MAX_SUBSCRIBER */
+
+/**
 * Abstraction of attribute associated with a resource.
 */
 struct coap_attr_t {


src/net.c
```diff
@@ -2916,6 +2916,9 @@ handle_request(coap_context_t *context, coap_session_t *session, coap_pdu_t *pdu
                                      coap_encode_var_safe(buf, sizeof (buf),
                                                           resource->observe),
                                      buf);
+          } else {
+            resp = 409;
```
mrdeep1, line 2920: This is an invalid response code for setting up an Observe registration.  As per [RFC7641 4.1. Request](https://datatracker.ietf.org/doc/html/rfc7641#section-4.1)
````
   A server that is unable or unwilling to add a new entry to the list
   of observers of a resource MAY silently ignore the registration
   request and process the GET request as usual.  The resulting response
   MUST NOT include an Observe Option, the absence of which signals to
   the client that it will not be notified of changes to the resource
   and, e.g., needs to poll the resource for its state instead.
````
which is how this should be processed.

File: src/resource.c [hunk 2]
```diff
@@ -735,6 +735,15 @@ static const uint16_t cache_ignore_options[] = { COAP_OPTION_ETAG };
 
   assert( session );
 
+  /* Check if there is already maximum number or subscribers present */
+#if (COAP_RESOURCE_MAX_SUBSCRIBER > 0)
+  uint8_t subscriber_count = 0;
```
mrdeep1, line 740: It is possible that you may get an unexpected wrap around here (if, say, COAP_RESOURCE_MAX_SUBSCRIBER was 256 or larger).  I would increase the size of this variable.

src/resource.c
```diff
@@ -735,6 +735,15 @@ static const uint16_t cache_ignore_options[] = { COAP_OPTION_ETAG };
...
+  LL_COUNT(resource->subscribers, s, subscriber_count);
+  if (subscriber_count > COAP_RESOURCE_MAX_SUBSCRIBER) {
+    return NULL; /* Signal error */
```
mrdeep1, line 743: What happens if a client is already registered and wants to check the subscription is already there for doing a refresh (i.e. client has just rebooted)?

I think this code should be a few lines further down after the `coap_find_observer()` test.

mrdeep1 (changes requested):
I like the principal behind this patch.

src/net.c
mrdeep1: It is possible to use a 5.03 code here [RFC7252 Section 5.9.3.4 Service Unavailable](https://www.rfc-editor.org/rfc/rfc7252#section-5.9.3.4) (with a Max-Age option), but this then means that the client has to repeat the request without Observe option register set to get the current resource value.  I think that deleting lines 2919-2921 is the correct way to go as the client will not see the Observe option in the response and know that the Observe was not registered.

include/coap3/resource.h
```diff
@@ -22,6 +22,10 @@
 #define COAP_RESOURCE_CHECK_TIME 2
 #endif /* COAP_RESOURCE_CHECK_TIME */
 
+#ifndef COAP_RESOURCE_MAX_SUBSCRIBER
+#define COAP_RESOURCE_MAX_SUBSCRIBER 0 /* There is no maximum */
+#endif /* COAP_RESOURCE_MAX_SUBSCRIBER */
+
```
mrdeep1, line 28: In general, variable definitions such as this which applications use should be in the publicly available header files.  However, in  this case the definition is used to determine how libcoap is compiled which the application is unable to change, so should not be visible to the applications.

Therefore, variable definitions that are libcoap library specific should be in the appropriate i`nclude/coap3/coap_*_include.h` file - in this case `include/coap3/coap_resource_internal.h`.  Please move this definition there.

mrdeep1 (changes requested):

rretanubun:
@mrdeep1 : Thank you for all the instructive feedback and clear change request. Changes applied as requested.

mrdeep1:
If you could squash the 4 commits into 1 and update the resultant commit description to reflect the current situation, that would be appreciated.

src/resource.c
```diff
@@ -763,6 +763,15 @@ static const uint16_t cache_ignore_options[] = { COAP_OPTION_ETAG };
     return s;
   }
 
+  /* Check if there is already maximum number or subscribers present */
```
mrdeep1, line 766: s/ or / of /

mrdeep1 (approved):

mrdeep1:
Thank you for the changes.

PR #1030, opened 2023-02-23
debug.c: define _GNU_SOURCE before including any header

mniestroj:
This is a fragment of feature_test_macros(7) man pages:

>  NOTE: In order to be effective, a feature test macro must be defined
>  before including any header files. This can be done either in the com‐
>  pilation command (cc -DMACRO=value) or by defining the macro within the
>  source code before including any headers. The requirement that the macro
>  must be defined before including any header file exists because header
>  files may freely include one another.

Just follow that guideline, as that is necessary in an effort to port
libcoap to Zephyr RTOS (where defining _GNU_SOURCE after first includes
breaks project build).

### DIFF
--- a/src/coap_debug.c
+++ b/src/coap_debug.c
@@ -13,11 +13,11 @@
  * @brief Debug utilities
  */
 
-#include "coap3/coap_internal.h"
-
-#if defined(HAVE_STRNLEN) && defined(__GNUC__) && !defined(_GNU_SOURCE)
-#define _GNU_SOURCE 1
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
 #endif
+
+#include "coap3/coap_internal.h"
 
 #include <stdarg.h>
 #include <stdio.h>


mrdeep1:
I'm not entirely sure that we still need to define _GNU_SOURCE anymore - change introduced in 2015 961b5e0719d1bc34cdc868a53b4d13b57d049628 to fix a compiler warning which I am unable to reproduce.

Does Zephyr RTOS need this defined?

mniestroj:
> Does Zephyr RTOS need this defined?

Actually it is newlib from Zephyr SDK (which bundles toolchain) that needs that when building as part of Zephyr RTOS. This is part of `/opt/zephyr-sdk-0.15.1/arm-zephyr-eabi/arm-zephyr-eabi/include/sys/types.h`:

```
#if __BSD_VISIBLE
#include <machine/endian.h>
#include <sys/select.h>
#  define	physadr		physadr_t
#  define	quad		quad_t

#ifndef _IN_ADDR_T_DECLARED
typedef	__uint32_t	in_addr_t;	/* base type for internet address */
#define	_IN_ADDR_T_DECLARED
#endif

#ifndef _IN_PORT_T_DECLARED
typedef	__uint16_t	in_port_t;
#define	_IN_PORT_T_DECLARED
#endif
```

So by defining `_GNU_SOURCE` above definition of `in_port_t` is provided. Otherwise I get this error:
```
/libcoap/src/coap_debug.c: In function 'coap_print_addr':
/libcoap/src/coap_debug.c:189:3: error: unknown type name 'in_port_t'; did you mean 'io_port_t'?
  189 |   in_port_t port;
      |   ^~~~~~~~~
      |   io_port_t
```

mrdeep1:
Hmm, as that is the only place that uses `in_port_t`, if that is replaced with `uint16_t`, do you get any other errors?

mniestroj:
> Hmm, as that is the only place that uses `in_port_t`, if that is replaced with `uint16_t`, do you get any other errors?

After replacing with `uint16_t` it builds fine both with existing code (leaving `_GNU_SOURCE` as it is) or when removing `_GNU_SOURCE` entirely.

mrdeep1:
Thanks.  I agree that things should be organized differently based on `feature_test_macros(7)`, but the simple replacement of `in_port_t` with `uint16_t `is the safest way to move forward at this time.  If you can update this PR (and commit message) according, that would be great and I can get it merged.

mniestroj:
@mrdeep1 Created a separate PR (#1031), just to be clear about what we discussed and for future reference.

PR #1033, opened 2023-02-24
mutex: add COAP_MUTEX_DEFINE() and support Zepyr RTOS mutex

mniestroj:
Introduce new macro, which will be used for defining static mutex
objects, for use from COAP_CONSTRAINED_STACK enabled code. This macro
will automatically initialize mutex using platform/OS specific macro if
such exists.

Provide mutex implementation for Zephyr RTOS, so that libcoap inside Zephyr
can be used with COAP_CONSTRAINED_STACK enabled.

This PR is a minimal set of changes in-tree that allows to integrate `libcoap`
into Zephyr RTOS application. I/O, (D)TLS and time routines implementation can
be easily added out-of-tree, so that Zephyr socket layer (either UDP/TCP or
Zephyr specific DTLS/TLS) can be utilized. For reference, this work is for now
part of https://github.com/golioth/golioth-firmware-sdk/pull/76. There will be
possiblity to upstream libcoap integration with Zephyr in the future.

### DIFF
--- a/include/coap3/coap_mutex_internal.h
+++ b/include/coap3/coap_mutex_internal.h
@@ -70,7 +70,17 @@ typedef int coap_mutex_t;
 #define coap_mutex_trylock(a) *(a) = 1
 #define coap_mutex_unlock(a) *(a) = 0
 
-#else /* !WITH_CONTIKI && !WITH_LWIP && !RIOT_VERSION && !HAVE_PTHREAD_H && !HAVE_PTHREAD_MUTEX_LOCK */
+#elif defined(__ZEPHYR__)
+#include <zephyr/sys/mutex.h>
+
+typedef struct k_mutex coap_mutex_t;
+#define COAP_MUTEX_DEFINE(_name)			\
+	static SYS_MUTEX_DEFINE(_name)
+#define coap_mutex_lock(a) sys_mutex_lock(a, K_FOREVER)
+#define coap_mutex_trylock(a) sys_mutex_lock(a, K_NO_WAIT)
+#define coap_mutex_unlock(a) sys_mutex_unlock(a)
+
+#else /* !__ZEPYR__ && !WITH_CONTIKI && !WITH_LWIP && !RIOT_VERSION && !HAVE_PTHREAD_H && !HAVE_PTHREAD_MUTEX_LOCK */
 /* define stub mutex functions */
 #warning "stub mutex functions"
 typedef int coap_mutex_t;
--- a/include/coap3/coap_mutex_internal.h
+++ b/include/coap3/coap_mutex_internal.h
@@ -29,7 +29,8 @@
 #include <pthread.h>
 
 typedef pthread_mutex_t coap_mutex_t;
-#define COAP_MUTEX_INITIALIZER PTHREAD_MUTEX_INITIALIZER
+#define COAP_MUTEX_DEFINE(_name)			\
+	static coap_mutex_t _name = PTHREAD_MUTEX_INITIALIZER
 #define coap_mutex_lock(a) pthread_mutex_lock(a)
 #define coap_mutex_trylock(a) pthread_mutex_trylock(a)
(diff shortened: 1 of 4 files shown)


include/coap3/coap_mutex_internal.h
```diff
@@ -50,7 +52,8 @@ typedef mutex_t coap_mutex_t;
 #if NO_SYS
 /* Single threaded, no-op'd in lwip/sys.h */
 typedef int coap_mutex_t;
-#define COAP_MUTEX_INITIALIZER 0
+#define COAP_MUTEX_DEFINE(_name)			\
+	static coap_mutex_t _name
```
mrdeep1, line 56: So that it is properly initialized, I think I would like to see this as
```
    static coap_mutex_t _name = 0
````
even though it does not really make any difference.

File: include/coap3/coap_mutex_internal.h [hunk 2]
```diff
@@ -61,16 +64,28 @@ typedef int coap_mutex_t;
 #elif defined(WITH_CONTIKI)
 /* Contiki does not have a mutex API, used as single thread */
 typedef int coap_mutex_t;
-#define COAP_MUTEX_INITIALIZER 0
+#define COAP_MUTEX_DEFINE(_name)			\
+	static coap_mutex_t _name
```
mrdeep1, line 68: So that it is properly initialized, I think I would like to see this as
```
    static coap_mutex_t _name = 0
````
even though it does not really make any difference.

File: include/coap3/coap_mutex_internal.h [hunk 3]
```diff
@@ -61,16 +64,28 @@ typedef int coap_mutex_t;
...
+#define coap_mutex_trylock(a) sys_mutex_lock(a, K_NO_WAIT)
+#define coap_mutex_unlock(a) sys_mutex_unlock(a)
+
+#else /* !__ZEPYR__ && !WITH_CONTIKI && !WITH_LWIP && !RIOT_VERSION && !HAVE_PTHREAD_H && !HAVE_PTHREAD_MUTEX_LOCK */
 /* define stub mutex functions */
 #warning "stub mutex functions"
 typedef int coap_mutex_t;
-#define COAP_MUTEX_INITIALIZER 0
+#define COAP_MUTEX_DEFINE(_name)			\
+	static coap_mutex_t _name
```
mrdeep1, line 88: So that it is properly initialized, I think I would like to see this as
```
    static coap_mutex_t _name = 0
````
even though it does not really make any difference.

mrdeep1 (commented):
In principal looks fine, apart from a couple of very minor nits.

include/coap3/coap_mutex_internal.h
mniestroj: Since it is `static`, it will be automatically initialized to 0. I think that some compilers (some gcc version really, I don't use anything else) even generate a warning when a `static` is "unnecessary" (aka explicitly) initialized to 0. This is true for sure when `static` variable is defined outside of function body, but not sure about defining it inside function body (in that case there might be no warning). But since this macro could potentially be used outside of function body in the future, I think it would be better to **not** initialize it.

Do you still want to explicitly initialize to 0 in that case? :)

mniestroj: Well, above (warning generation by gcc) it true for `int` type, it might be different for `typedef` such as `coap_mutex_t`.

mrdeep1: Need to checkout for a few hours, will look at it when I get back

mrdeep1: Agreed, this initialization should not be done.

mrdeep1: Agreed, this initialization should not be done.

include/coap3/coap_mutex_internal.h
mrdeep1: Agreed, this initialization should not be done.

PR #1034, opened 2023-02-24
net: fix coap_log_warn() in coap_send()

mniestroj:
`pdu->actual_token.length` is a `size_t`, so use `"%zu"` format specifier.

`session->max_token_size` is a `uint32_t`, so use `"%" PRIu32` format specifier.

This allows to correctly handle those in a cross platform way.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -1142,7 +1142,7 @@ coap_send(coap_session_t *session, coap_pdu_t *pdu) {
   if (COAP_PDU_IS_REQUEST(pdu) &&
       pdu->actual_token.length > session->max_token_size) {
     coap_log_warn(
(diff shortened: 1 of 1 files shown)


mrdeep1:
There are several other places where there is an issue with `uint32_t `printf frmatting, and so the use of `PRIu32 `from `<inttypes.h>` has started to be used in various places.

Perhaps this would the right direction to go here (with `#include <inttypes.h>` moved to `coap_internal.h`).

mrdeep1:
`inttypes.h` is now included in all of the libcoap library - as per merged #1035.  Please use `PRIu32 `instead for `max_token_size `printing.

PR #1045, opened 2023-03-03
coap-client: Make retransmission give up more visible

rretanubun:
Useful when coap-client is used to send block transfers.

Without this, users have to run coap-client with `-v 7` before they can see why coap-client ends earlier than expected.

Logging at warning level to match the default log level set by coap-client source code.


mrdeep1:
Looks good to me - thanks.

Issue #757, opened 2021-10-05
Proposal to remove COAP_MEDIATYPE_ANY


obgm:
[COAP_MEDIATYPE_ANY](https://github.com/obgm/libcoap/blob/c190c4960fd88f1efaf3468a09be82052649408b/include/coap3/pdu.h#L226) has been defined as `0xff` in the very early implementation phase of libcoap (i.e., mid 2010) to gain a bit of flexibility with media type handling (what later has become *Content-Format* in RFC 7252). It was originally intended to mimic the HTTP `Accept: *` but the CoAP Accept option then was defined to work in a different way. 
**A major issue with this definition is that the value 255 is in no way special and therefore in principle is open for IANA registration.**
Since `COAP_MEDIATYPE_ANY` is not used in the libcoap core and most likely not in applications as there is little use of this in the public libcoap API, my feeling is that this definition should be removed. (In theory, any value above 64999 would do, even, e.g., `0x1ffff` but keeping unused code clutters the API).

mrdeep1:
Given that if the CoAP Accept Option is not present in the request, this indicates 'Any' ([Accept](https://datatracker.ietf.org/doc/html/rfc7252#section-5.10.4)) there is no need for the non-standard COAP_MEDIATYPE_ANY (unlikely to be recognized by any other CoAP implementation) to be used in an Accept Option.

I have no issues with it being deleted.

PR #1095, opened 2023-04-24
Remove COAP_MEDIATYPE_ANY

obgm:
As discussed earlier (#757), the Content-Format `COAP_MEDIATYPE_ANY`  now is removed from libcoap.


mrdeep1:
What code version does the tinydtls change refer to?

obgm:
Uh, there seems to have some garbage sneaked in. The only change would be the removal of the `#define` in `pdu.h`.

PR #1159, opened 2023-07-05
Fix for piggybacked response answering request with No-Response

magdalenaszumny:
Fix for piggybacked response answering to request with No-Response option set.

When No-Response option is interpreted in no_response function from net.c, for piggybacked response payload is removed to send only bare ACK. That means that used_size is set to 0. This causes an error in coap_get_data_large function from pdu.c, because when used_size is 0 then variable len has incorrect (and very large) value. This causes application to crash.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -2543,6 +2543,7 @@ no_response(coap_pdu_t *request, coap_pdu_t *response,
           response->actual_token.length = 0;
           response->e_token_length = 0;
           response->used_size = 0;
+          response->data = NULL;
           return RESPONSE_SEND;
         }
         else {


mrdeep1:
Thanks for looking into this issue and coming up with a solution.  While you have come up with a safe solution, I wonder if the same sort of issue is elsewhere.  I think a better solution would be
````
diff --git a/src/net.c b/src/net.c
index 8a2d257..28f73ce 100644
--- a/src/net.c
+++ b/src/net.c
@@ -2543,6 +2543,7 @@ no_response(coap_pdu_t *request, coap_pdu_t *response,
           response->actual_token.length = 0;
           response->e_token_length = 0;
           response->used_size = 0;
+          response->data = NULL;
           return RESPONSE_SEND;
         }
         else {
````
and so `coap_get_data_large()` will catch this and return just before your suggested change.  It means that `->data` is not mis-interpreted elsewhere.  It would be good to check that this solves your particular issue.

magdalenaszumny:
> 

I think it is very good idea! Thank you for suggestion. I implemented it as you suggested. I checked and it solves the issue I noticed in my setup.

mrdeep1:
Thanks for checking this out.  If you could squash the 2 commits into 1, and give the single commit a more meaningful title such as
````
net.c: Clear data offset for piggybacked response answering request with No_response
````

magdalenaszumny:
> Thanks for checking this out. If you could squash the 2 commits into 1, and give the single commit a more meaningful title such as
> 
> ```
> net.c: Clear data offset for piggybacked response answering request with No_response
> ```

Done :)

PR #1175, opened 2023-07-16
Fix potential index out of bounds

mjunix:


mrdeep1:
Thanks for finding this, debugging it and creating a patch.

PR #1201, opened 2023-08-04
Contiki-NG: Switch to secure PRNG

kkrentz:
Contiki-NG now has a secure PRNG, which we can use in `coap_prng.c`.


mrdeep1:
You need to add in `csprng_feed()` to `coap_prng_init()`.

Should we be using `release/v4.9` or `develop/v4.9` as a known supported version?

kkrentz:
`release/v4.9` works. Contiki-NG takes care of seeding its insecure and secure PRNG while booting.

PR #1211, opened 2023-08-16
coap_block.c: Use correct format specifier when logging chunk size

kkrentz:
Two unexact format specifiers in `coap_block.c` caused compiler errors on an embedded target. This PR fixes the issue.


mrdeep1:
Thanks for finding and providing a fix for these.

PR #1216, opened 2023-08-30
fix memory leak in coap_async

osobiehl:
This is a bug I encountered a while ago while ago.

LL_PREPEND was setting s as the first element in the linked list, and then memset was clearing it; resulting in any other pending async requests being lost


mrdeep1:
Thank you for finding and fixing this.

PR #1218, opened 2023-09-02
coap_address.c: do not enable IPV4 broadcast as multicast with Zephyr

### LINKED ISSUE #1217: Build issue with ifaddr.h
## Environment

- Build System:             CMake (via Zephyr `west`)
- Operating System:         Linux
- Operating System Version: Ubuntu 24.04
- Hosted Environment:       Zephyr

## libcoap Configuration Summary

Copy the "libcoap Configuration Summary" output lines here after running
`./configure` (with any options that you have specified)
  or
`cmake ..` (with any -D options that you have specified)
  or if neither of the above two methods is used (do from within the libcoap directory)
`git describe --tags`

## Problem Description

Seeing the following build error:
```
/__w/golioth-firmware-sdk/golioth-firmware-sdk/modules/lib/golioth-firmware-sdk/external/libcoap/src/coap_address.c:160:10: fatal error: ifaddrs.h: No such file or directory
  160 | #include <ifaddrs.h>
      |          ^~~~~~~~~~~
```
Which I believe is due to the fact that `ifaddrs.h` is not included in the Zephyr [minimal libc](https://docs.zephyrproject.org/latest/develop/languages/c/minimal_libc.html).

I can see that this addition was wrapped to be excluded when ESP IDF is being used -- perhaps a similar conditional for Zephyr could be added. I am happy to implement if so.

### Expected Behavior

I expected for compilation to be successful.

### Actual Behavior

Compilation failed with steps described above.

### Steps to reproduce

A full reproduction of the build can be found [here](https://github.com/golioth/golioth-firmware-sdk/actions/runs/6054900354/job/16433030414?pr=163).

hasheddan:
Adds a check to ensure IPV4 broadcast as multicast is not enabled when using Zephyr as it will break builds with Zephyr's minimal libc, which is missing ifaddr.h.

Fixes #1217

### DIFF
--- a/src/coap_address.c
+++ b/src/coap_address.c
@@ -139,7 +139,7 @@ coap_is_mcast(const coap_address_t *a) {
   return 0;
 }
 
-#if !defined(WIN32)
+#if !defined(WIN32) && !defined(__ZEPHYR__)
 
 #ifndef COAP_BCST_CNT
 #define COAP_BCST_CNT 15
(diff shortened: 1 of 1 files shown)


mrdeep1:
I would like to make the fix more generic (i.e. other build environments that don't have ifaddr.h), but this is a safe fix for Zephyr.

Issue #1225, opened 2023-09-08
Do not destroy mutexes in `coap_cleanup()`


hasheddan:
## Environment

- Build System:             CMake
- Operating System:        Ubuntu
- Operating System Version: 20.04
- Hosted Environment:      ESP-IDF

## libcoap Configuration Summary

Configuration can be found here: https://github.com/golioth/golioth-firmware-sdk/pull/163/files#diff-152d8027819ee12f017ab0edd01f2b2f218d24b7f751070b7d130842be298a3a

## Problem Description

Tests run [here](https://github.com/golioth/golioth-firmware-sdk/tree/main/examples/esp_idf/test) reveal a [null queue when acquiring a semaphore](https://github.com/golioth/golioth-firmware-sdk/actions/runs/6115076185/job/16598003472#step:9:443) (see assert [here](https://github.com/espressif/esp-idf/blob/62ee4135e033cc85eb0d7572e5b5d147bcb4349e/components/freertos/FreeRTOS-Kernel-V10.5.1/queue.c#L1586)). The decoded stack trace looks as follows:
```
0x4037b36f: xt_utils_compare_and_set at /home/hasheddan/code/github.com/espressif/esp-idf/components/xtensa/include/xt_utils.h:215
 (inlined by) esp_cpu_compare_and_set at /home/hasheddan/code/github.com/espressif/esp-idf/components/esp_hw_support/cpu.c:483
0x40380bf9: spinlock_acquire at /home/hasheddan/code/github.com/espressif/esp-idf/components/esp_hw_support/include/spinlock.h:121
 (inlined by) xPortEnterCriticalTimeout at /home/hasheddan/code/github.com/espressif/esp-idf/components/freertos/FreeRTOS-Kernel/portable/xtensa/port.c:501
0x4037e4b4: vPortEnterCritical at /home/hasheddan/code/github.com/espressif/esp-idf/components/freertos/FreeRTOS-Kernel/portable/xtensa/include/freertos/portmacro.h:584
 (inlined by) xQueueSemaphoreTake at /home/hasheddan/code/github.com/espressif/esp-idf/components/freertos/FreeRTOS-Kernel/queue.c:1671
0x4200390d: pthread_mutex_lock_internal at /home/hasheddan/code/github.com/espressif/esp-idf/components/pthread/pthread.c:614
0x42003a9a: pthread_mutex_lock at /home/hasheddan/code/github.com/espressif/esp-idf/components/pthread/pthread.c:644
[...]

PR #1226, opened 2023-09-08
Do not destroy mutexes in coap_cleanup()

hasheddan:
If mutexes are destroyed in coap_cleanup() they can never be re-initialized as calling coap_startup() short-circuits after its first invocation. This updates to not destroy the mutexes on calls to coap_cleanup().

Fixes https://github.com/obgm/libcoap/issues/1225

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -4043,15 +4043,6 @@ coap_cleanup(void) {
   coap_stop_io_process();
 #endif
   coap_dtls_shutdown();
-
-#if COAP_CONSTRAINED_STACK
-  coap_mutex_destroy(&m_show_pdu);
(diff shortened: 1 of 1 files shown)


mrdeep1:
`coap_cleanup()` also destroys things like the DTLS environment as well as Windows / Contiki environment setup, so currently it is unsafe to assume libcoap is stable in any form after calling `coap_cleanup()`. See [coap_cleanup(3)](https://libcoap.net/doc/reference/develop/man_coap_cleanup.html).

If the mutexes allocate any memory in `coap_startup()`, then there will be a memory leak if they are not cleared down in `coap_cleanup()`.

The issue you are finding here needs to be solved in another way.  See some of my suggestions in #1225.

PR #1233, opened 2023-09-14
Contiki-NG: Fix compilation on macOS

kkrentz:
This PR fixes that Contiki-NG fails to compile on macOS due to an unused function and a redefinition of `ntohs`.


mrdeep1 (commented):
Thanks for working on this.

At some point, all the commits need to be squashed into a single commit with appropriate documentation.

PR #1241, opened 2023-09-27
Contiki-NG: Fix retransmissions and simplify cleanup

kkrentz:
I noticed that retransmissions are sometimes sent only after receiving something. The reason was that no `tcpip_event` occurred after sending, which would have caused the retransmission timer to be re-scheduled. This PR fixes the issue. Besides, I added a commit that leverages Contiki-NG's `process_exit` function for simplifying the cleanup procedure in `coap_io_contiki.c`.

### DIFF
--- a/include/coap3/coap_io_internal.h
+++ b/include/coap3/coap_io_internal.h
@@ -147,12 +147,13 @@ void coap_epoll_ctl_add(coap_socket_t *sock, uint32_t events, const char *func);
 void coap_epoll_ctl_mod(coap_socket_t *sock, uint32_t events, const char *func);
 
 /**
- * Update the epoll timer fd as to when it is to trigger.
+ * Update when to continue with I/O processing, unless packets come in in the
+ * meantime. Typically, this timeout triggers retransmissions.
  *
- * @param context The context to update the epoll timer on.
- * @param delay The time to delay before the epoll timer fires.
+ * @param context The CoAP context.
+ * @param delay The time to delay before continuing with I/O processing.
  */
-void coap_update_epoll_timer(coap_context_t *context, coap_tick_t delay);
+void coap_update_io_timer(coap_context_t *context, coap_tick_t delay);
 
 #ifdef WITH_LWIP
 ssize_t coap_socket_send_pdu(coap_socket_t *sock, coap_session_t *session,
--- a/src/coap_io_contiki.c
+++ b/src/coap_io_contiki.c
@@ -31,7 +31,7 @@ coap_stop_io_process(void) {
 }
 
 static void
-on_prepare_timer_expired(void *ptr) {
+on_io_timer_expired(void *ptr) {
   PROCESS_CONTEXT_BEGIN(&libcoap_io_process);
   prepare_io((coap_context_t *)ptr);
   PROCESS_CONTEXT_END(&libcoap_io_process);
@@ -39,15 +39,15 @@ on_prepare_timer_expired(void *ptr) {
 
 void
 coap_update_io_timer(coap_context_t *ctx, coap_tick_t delay) {
-  if (!ctimer_expired(&ctx->prepare_timer)) {
-    ctimer_stop(&ctx->prepare_timer);
+  if (!ctimer_expired(&ctx->io_timer)) {
+    ctimer_stop(&ctx->io_timer);
   }
   if (!delay) {
     process_post(&libcoap_io_process, PROCESS_EVENT_POLL, ctx);
   } else {
-    ctimer_set(&ctx->prepare_timer,
+    ctimer_set(&ctx->io_timer,
                CLOCK_SECOND * delay / 1000,
-               on_prepare_timer_expired,
+               on_io_timer_expired,
                ctx);
   }
 }
@@ -58,12 +58,12 @@ prepare_io(coap_context_t *ctx) {
   coap_socket_t *sockets[1];
   static const unsigned int max_sockets = sizeof(sockets)/sizeof(sockets[0]);
   unsigned int num_sockets;
-  unsigned timeout;
+  unsigned next_io;
 
   coap_ticks(&now);
-  timeout = coap_io_prepare_io(ctx, sockets, max_sockets, &num_sockets, now);
-  if (timeout) {
-    coap_update_io_timer(ctx, timeout);
+  next_io = coap_io_prepare_io(ctx, sockets, max_sockets, &num_sockets, now);
+  if (next_io) {
+    coap_update_io_timer(ctx, next_io);
   }
 }
 
--- a/src/coap_io_contiki.c
+++ b/src/coap_io_contiki.c
@@ -18,18 +18,15 @@
 #include "contiki-net.h"
 
 PROCESS(libcoap_io_process, "libcoap I/O");
-static int was_io_process_stopped;
 
 void
 coap_start_io_process(void) {
-  was_io_process_stopped = 0;
   process_start(&libcoap_io_process, NULL);
 }
 
 void
 coap_stop_io_process(void) {
-  was_io_process_stopped = 1;
-  process_poll(&libcoap_io_process);
+  process_exit(&libcoap_io_process);
 }
 
 static void
@@ -56,12 +53,11 @@ on_prepare_timer_expired(void *ptr) {
 PROCESS_THREAD(libcoap_io_process, ev, data) {
   coap_socket_t *coap_socket;
 
+  PROCESS_EXITHANDLER(goto exit);
   PROCESS_BEGIN();
-  while (!was_io_process_stopped) {
+
+  while (1) {
     PROCESS_WAIT_EVENT();
-    if (was_io_process_stopped) {
-      break;
-    }
     if (ev == tcpip_event) {
       coap_socket = (struct coap_socket_t *)data;
       if (!coap_socket) {
@@ -76,6 +72,8 @@ PROCESS_THREAD(libcoap_io_process, ev, data) {
       }
     }
   }
+exit:
+  coap_log_info("libcoap_io_process: stopping\n");
   PROCESS_END();
 }
 
--- a/src/coap_io_contiki.c
+++ b/src/coap_io_contiki.c
@@ -56,7 +56,7 @@ static void
 prepare_io(coap_context_t *ctx) {
   coap_tick_t now;
   coap_socket_t *sockets[1];
-  unsigned int max_sockets = sizeof(sockets)/sizeof(sockets[0]);
+  static const unsigned int max_sockets = sizeof(sockets)/sizeof(sockets[0]);
   unsigned int num_sockets;
   unsigned timeout;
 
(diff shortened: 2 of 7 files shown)


mrdeep1:
I am not convinced this is the way to go - in particular Observe is not triggering in the example Contiki-NG server.

As I read the code, there are multiple threads of interest, which we need to be careful about overlapping activity.

`on_prepare_timer_expired(`) needs to be invoked, and hence `coap_io_prepare_io()` whenever `coap_resource_notify_observers()` is invoked (or immediately afterwards) by the `coap_server_process` thread.  For safety, `coap_io_prepare_io()` should only be run by the `libcoap_io_process` thread.

(Contiki version of) coap_io_process() is invoked by the `libcoap_io_process` thread which is fine, but can be called by an app directly, or indirectly by `coap_client_delay_first()` when a client app tries to send the first packet.  So, I think the current logic of `coap_io_process()` should be copied into the `libcoap_io_process` thread, and then a new `coap_io_process()` should do a `process_push()` (or `process_poll()`) to trigger the `libcoap_io_process` thread.

However, then the new `coap_io_process()` needs to not immediately return, but delay until a new tcpip event occurs, or a timer expires based on the `timeout_ms `value whichever comes first.  

Note: In examples/contiki/server.c, the `etimer_set()` should read (to get time updating once per second)
````
  etimer_set(&dirty_timer, 1 * CLOCK_SECOND);
````

My initial tests show that `process_event_t on_sent` patch is not needed if the above is followed (but have not fixed the delay for the new `coap_io_process()` yet).

 What do you think?

**Comment by mrdeep1** [comment]
With just these changes
````
diff --git a/examples/coap-client.c b/examples/coap-client.c
index 18b6777f..bdcc79da 100644
--- a/examples/coap-client.c
+++ b/examples/coap-client.c
@@ -1907,6 +1907,7 @@ main(int argc, char **argv) {
     quit = 1;
   }
   repeat_count--;
+sleep (30);

   while (!quit &&                /* immediate quit not required .. and .. */
          (tracked_tokens_count || /* token not responded to or still observe */
diff --git a/examples/contiki/server.c b/examples/contiki/server.c
index b10be59f..2055958e 100644
--- a/examples/contiki/server.c
+++ b/examples/contiki/server.c
@@ -176,12 +176,13 @@ PROCESS_THREAD(coap_server_process, ev, data) {
   init_coap_resources(coap_context);

   /* etimer_set(&notify_timer, 5 * CLOCK_SECOND); */
-  etimer_set(&dirty_timer, 30 * CLOCK_SECOND);
+  etimer_set(&dirty_timer, 1 * CLOCK_SECOND);

   while (1) {
     PROCESS_YIELD();
     if (ev == PROCESS_EVENT_TIMER && etimer_expired(&dirty_timer)) {
       coap_resource_notify_observers(time_resource, NULL);
+      coap_io_process(coap_context, 0);
       etimer_reset(&dirty_timer);
     }
   }
````
building both examples/coap-client and (in examples/c.ontiki) server.native

Running server.native under valgrind, there are a lot of complaints when running the (temporary fix) examples coap-client and it hits the sleep statement to force and observe re-transmit.
````
examples/coap-client coap://[FD00:0000:0000:0000:0302:0304:0506:0708]/time -s6  -w
````
My suspicion is with
````
+  PROCESS_CONTEXT_BEGIN(&libcoap_io_process);
+  prepare_io((coap_context_t *)ptr);
+  PROCESS_CONTEXT_END(&libcoap_io_process);
````
messing up some stack usage.

This change needs to be made (to handle NULL pointers)
````
diff --git a/src/coap_mem.c b/src/coap_mem.c
index 7bc2cbb1..a0c7d1aa 100644
--- a/src/coap_mem.c
+++ b/src/coap_mem.c
@@ -500,7 +500,8 @@ coap_realloc_type(coap_memory_tag_t type, void *p, size_t size) {

 void
 coap_free_type(coap_memory_tag_t type, void *ptr) {
-  heapmem_free(ptr);
+  if (ptr)
+    heapmem_free(ptr);
 }
 #endif /* WITH_CONTIKI */

````

**Comment by kkrentz** [comment]
Thank you for pointing me to these other "threads of interests". Conveniently, `coap_update_epoll_timer` seems to get called whenever an update is necessary already. Building on that, I introduced a more general function `coap_update_timer`.

I do not know why Valgrind complained about `PROCESS_CONTEXT_BEGIN`, but I now take the detour via `process_poll`.

Actually, I realized that calling `process_exit` is not only a simplification, but also avoids conflicts when `coap_startup` is called right after `coap_cleanup`. With `process_exit`, exiting the `libcoap_io_process` happens instantly. So, no conflicts can happen.

**Comment by mrdeep1** [comment]
It has taken me some time, but I have now got to the bottom of why `valgrind` was complaining when outputting debug info in `coap_wait_ack()` as `coap_prng()` called in `coap_send_internal()` was not doing anything, and so the call to `coap_calc_timeout()` in `coap_send_internal()` was putting uninitialized information into `node->timeout`.

`csprng_feed()` is not called, and so any call to `csprng_rand()` does not update the `buf` parameter, and so what that points to is uninitialized.

A side affect of this not calling `csprng_feed()` is that any call to `coap_prng()` was not creating any randomness other that what was on the stack at that time where `buf` was located.

[csprng.patch](https://github.com/obgm/libcoap/files/12794864/csprng.patch) appears to work for me.

**Comment by kkrentz** [comment]
- As for `csprng_feed()`, I think this should rather be addressed in Contiki-NG directly. I created a PR there: https://github.com/contiki-ng/contiki-ng/pull/2730.
- Equally, I created PR for suppressing the `heapmem_free` warning: https://github.com/contiki-ng/contiki-ng/pull/2729.
- `coap_client_delay_first` cannot be implemented via busy-waiting in `coap_io_process` on Contiki-NG as this would block all other protothreads. Typically, this will lead to watchdog reboots. For the time being, I limited the support for `coap_io_process`, which causes `coap_client_delay_first` to return an error. As this function is only called in `coap_send` when using `coap_config.h.contiki`, clients can work around this limitation by sending requests one after another.
- As for your patch, it seems unnecessary to add `coap_io_process(coap_context, 0);` to `server.c`. Now, `coap_resource_notify_observers` calls `coap_update_timer(r->context, 0);`, which polls the `libcoap_io_process`, which performs `coap_io_prepare_io`. Presumably, it is unnecessary to call `coap_io_process` in addition.

**Comment by mrdeep1** [comment]
> As for csprng_feed(), I think this should rather be addressed in Contiki-NG directly. I created a PR
> there: https://github.com/contiki-ng/contiki-ng/pull/2730.

I agree that is a long term solution which needs to be done for every possible platform.  Under `examples/contiki` when building Contiki, it pulls in `release/v4.9` - this will need to be changed at some point.  I would prefer libcoap to have a temporary fix until the proper fix is in mainstream Contiki-NG as `coap_wait_ack()` continues to get valgrind complaints.

> Equally, I created PR for suppressing the heapmem_free warning:
> https://github.com/contiki-ng/contiki-ng/pull/2729.

Similarly, we need in a temp fix until mainstream Contiki-NG is in place.  Save others the trouble of unnecessary debugging.

> coap_client_delay_first(): As this function is only called in coap_send when using coap_config.h.contiki

Agree that is is is invoked when COAP_OSCORE_SUPPORT, COAP_Q_BLOCK_SUPPORT are not set.  It is also called when extended tokens are being used and the client is testing whether the server supports this.

> As for your patch, it seems unnecessary to add coap_io_process(coap_context, 0); to server.c.

Agreed.

I see that `process_thread_coap_server_process()` gets called every 1 sec now in `server.c`, and so we are  setting up the libcoap information every time - which just needs to be the once, and certainly not multi-threaded safe.

I also think that `coap_io_process()` should just invoke` process_thread_libcoap_io_process(`) and not do `coap_io_do_io()` in case it is on a separate thread, but `coap_io_do_io()` should be done by `process_thread_libcoap_io_process()` instead of calling `coap_io_process()`.

Fixes for both in [contiki1.patch](https://github.com/obgm/libcoap/files/12803000/contiki1.patch).

**Comment by kkrentz** [comment]
- The CSPRNG should not be fed with an insecure seed as subsequent calls to `csprng_rand` would succeed, creating other security issues. Maybe, checking the return value of `coap_prng` is a better way out. This would overcome similar issues with other platforms and other random number generators.
- In `server.c`, the execution continues from `PROCESS_YIELD`. There is no need to jump over the initialization part.
- There seems to be no easy way to support `coap_client_delay_first` on Contiki-NG. Busy-waiting in `coap_io_process` prevents other protothreads from making progress, which is why `doing_first` will never turn false. Now, `coap_client_delay_first` at least reports an error back.

**Comment by mrdeep1** [comment]
> The CSPRNG should not be fed with an insecure seed as subsequent calls to csprng_rand would succeed,
> creating other security issues. Maybe, checking the return value of coap_prng is a better way out. This would
> overcome similar issues with other platforms and other random number generators.

Yes, perhaps the best way forward here is to not use csprng_rand() in libcoap.

> In server.c, the execution continues from PROCESS_YIELD. There is no need to jump over the initialization part.

I agree that it would appear that PROCESS_YIELD should just "continue", but I was unsure how `ev` got updated as it is passed in as a parameter, and then found by setting a breakpoint in gdb at the start of `process_thread_coap_server_process()`, this function is getting called once a second.

> There seems to be no easy way to support `coap_client_delay_first `on Contiki-NG. 

Agreed, cant think of anything at present, especially on finding how PROCESS_YIELD actually works.

**Comment by mrdeep1** [comment]
If you could update your develop branch and rebase against that, this would help to move this forward.

Please read https://github.com/obgm/libcoap/wiki/Unable-to-update-develop-branch-from-remote-repository before updating your develop branch.

File: src/coap_io_contiki.c [hunk 1]
```diff
@@ -217,7 +239,13 @@ int
...
   coap_tick_t before, now;
 
+  if (timeout_ms != COAP_IO_NO_WAIT) {
+    coap_log_err("coap_io_process must be called with COAP_IO_NO_WAIT\n");
+    return -1;
+  }
+
   coap_ticks(&before);
+  prepare_io(ctx);
   coap_io_do_io(ctx, before);
```
mrdeep1, line 249: I think these 2 lines should be wrapped with
````
PROCESS_CONTEXT_BEGIN(&libcoap_io_process);
PROCESS_CONTEXT_END(&libcoap_io_process);
````
so everything libcoap wise is run under the libcoap_io_process.

File: include/coap3/coap_io_internal.h [hunk 2]
```diff
@@ -147,12 +147,12 @@ void coap_epoll_ctl_add(coap_socket_t *sock, uint32_t events, const char *func);
...
- * Update the epoll timer fd as to when it is to trigger.
+ * Re-schedule when a call to coap_io_prepare_io is due.
  *
- * @param context The context to update the epoll timer on.
- * @param delay The time to delay before the epoll timer fires.
+ * @param context The context that requires the call.
+ * @param delay The time span until the call is due.
  */
-void coap_update_epoll_timer(coap_context_t *context, coap_tick_t delay);
+void coap_update_timer(coap_context_t *context, coap_tick_t delay);
```
mrdeep1, line 155: Perhaps a better function name is coap_update_refresh_timer() to give a clue to the reader as to what is happening here.  I do like the re-use of the coap_update_epoll_timer() concept to trigger the refresh of the libcoap logic when, say, a packet need to be re-transmitted.

mrdeep1 (commented):
Some minor comments.

PR #1265, opened 2023-10-19
Suppress RST when the processing of a non-empty ACK fails

kkrentz:
Replying an RST in response to an ACK is not compliant, is it? Hence, we should sort out this case when, e.g., processing a piggybacked response.

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -3406,8 +3406,9 @@ handle_response(coap_context_t *context, coap_session_t *session,
 
   /* Call application-specific response handler when available. */
   if (context->response_handler) {
-    if (context->response_handler(session, sent, rcvd,
-                                  rcvd->mid) == COAP_RESPONSE_FAIL)
+    if ((context->response_handler(session, sent, rcvd,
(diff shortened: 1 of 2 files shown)


src/coap_net.c
```diff
@@ -3406,8 +3406,9 @@ handle_response(coap_context_t *context, coap_session_t *session,
 
   /* Call application-specific response handler when available. */
   if (context->response_handler) {
-    if (context->response_handler(session, sent, rcvd,
-                                  rcvd->mid) == COAP_RESPONSE_FAIL)
+    if ((context->response_handler(session, sent, rcvd,
+                                   rcvd->mid) == COAP_RESPONSE_FAIL)
+        && (rcvd->type != COAP_MESSAGE_ACK))
```
mrdeep1, line 3411: My preference is to have the && at the end of the previous line (3410), which then means that the rest of line 3411 needs to be aligned under the opening (.

mrdeep1 (commented):
This change looks good - one minor nit.

PR #1275, opened 2023-11-06
OSCORE: Treat CoAP pings as requests

kkrentz:
When sending CoAP pings on an OSCORE-protected session, I run into `PDU could not be encrypted`. The cause is that OSCORE treats them as responses.


include/coap3/coap_pdu_internal.h
```diff
@@ -102,6 +102,8 @@
 #define COAP_PDU_IS_REQUEST(pdu)   (!COAP_PDU_IS_EMPTY(pdu) && (pdu)->code < 32)
 #define COAP_PDU_IS_RESPONSE(pdu)  ((pdu)->code >= 64 && (pdu)->code < 224)
 #define COAP_PDU_IS_SIGNALING(pdu) ((pdu)->code >= 224)
+#define COAP_PDU_IS_PING(pdu)      (COAP_PDU_IS_EMPTY(pdu) && \
+                                    ((pdu)->type == COAP_MESSAGE_CON))
```
mrdeep1, line 106: This needs to include `(pdu)->code == COAP_SIGNALING_CODE_PING` to test for when using TCP.

````
+#define COAP_PDU_IS_PING(pdu)      ((COAP_PDU_IS_EMPTY(pdu) && \
+                                     (pdu)->type == COAP_MESSAGE_CON) || \
+                                    (pdu)->code == COAP_SIGNALING_CODE_PING)
````

PR #1276, opened 2023-11-07
Fix inconsistent tokens received by other servers

cisiqo:
When use `coap-client` with -T command, the other server received inconsistent.
```
coap-client send token "aaa", the other server received "aab"
coap-client send token "obstoken", the other server received "obstokeo"
```

### DIFF
--- a/src/coap_session.c
+++ b/src/coap_session.c
@@ -1583,7 +1583,7 @@ void
 coap_session_new_token(coap_session_t *session, size_t *len,
                        uint8_t *data) {
   *len = coap_encode_var_safe8(data,
-                               sizeof(session->tx_token), ++session->tx_token);
+                               sizeof(session->tx_token), session->tx_token++);
 }
 
 uint16_t


mrdeep1:
Thanks for diagnosing, reporting and fixing this.

golemik98:
Hi!

I think this fix can cause problems in one particular scenario, which I stumbled  on lately.
https://github.com/obgm/libcoap/blob/c7c3275e5326396b0b41463f9205c1c84295d4ea/src/coap_net.c#L1421
When we use PLAIN COAP we do not call this code unless it is GET method with observation option.
When it is GET method with observation option we call this and set `state_token` as pre-incremented value of tx_token from session.
Then we want to send next request to the server with newly generated token, by this function: https://github.com/obgm/libcoap/blob/c7c3275e5326396b0b41463f9205c1c84295d4ea/src/coap_session.c#L1603
So the message token is (post)not incremented `tx_token`, which is set to `state_token`. Then server is responding with the same token (as it supposed to do). 
We handle this response in function https://github.com/obgm/libcoap/blob/c7c3275e5326396b0b41463f9205c1c84295d4ea/src/coap_block.c#L3469
and what is happening next, we encounter this condition: https://github.com/obgm/libcoap/blob/c7c3275e5326396b0b41463f9205c1c84295d4ea/src/coap_block.c#L3496
which happen this time to be false. It is false because `token_match != STATE_TOKEN_BASE(p->state_token)` match. So this function starts to malfunction, and its result is handling response of wrong (updated) token, not the token that was originally send (this token was generated with the same value as `state_token`) but due to this condition: https://github.com/obgm/libcoap/blob/c7c3275e5326396b0b41463f9205c1c84295d4ea/src/coap_block.c#L3797, updated.

This happens always when we use PLAIN COAP and send GET with observation and after that any other request from client, client is handling response with wrong (updated by itself) token. Problem is in my opinion with this bug-fix that you mention here  https://github.com/obgm/libcoap/pull/1276.
This case is not taking place when we do increment token to PDU first then use this value as newly generated token.
Would be great if you consider this issue. Thanks!

PR #1281, opened 2023-11-19
Contiki-NG: Crypto adapter for OSCORE

kkrentz:
This implements the crypto primitives required by OSCORE using Contiki-NG's potentially hardware-accelerated drivers. For this to work, some changes to Contiki-NG have to be merged first, namely https://github.com/contiki-ng/contiki-ng/pull/2779 and https://github.com/contiki-ng/contiki-ng/pull/1653.

### DIFF
--- a/include/coap3/coap_internal.h
+++ b/include/coap3/coap_internal.h
@@ -85,6 +85,11 @@
 /*
  * Include all the header files that are for internal use only.
  */
+#if COAP_OSCORE_SUPPORT && defined(WITH_CONTIKI)
+#include "lib/aes-128.h"
+#include "lib/ccm-star.h"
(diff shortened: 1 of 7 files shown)


mrdeep1:
In principal, having support for specific hardware acceleration makes sense.

However, I would prefer things to be wrapped with both WITH_CONTIKI and, say, CCM_STAR_CONF (if that is the correct one) in case a particular Contiki hardware platform cannot support CCM_STAR_CONF etc.

For OSCORE to work, is this dependent on the 2 Contiki-NG PRs?  If so, I would prefer to see those merged before we merge this PR otherwise others might get confused when things are not working as expected.

Issue #1306, opened 2024-01-02
Client fails to ACK a retransmitted CON message


pptz:
## Environment

- Build System:             cmake (via esp-idf SDK wrappers)
- Operating System:         MacOS
- Operating System Version: 13.5
- Hosted Environment:       ESP-IDF (code runs on an ESP32S3)

## libcoap Configuration Summary

I can provide details if needed, but I don't think anything other than me using an unreliable protocol (DTLS) to establish resource observation by a client matters here.
The version I observe the issue for is `v4.3.1-dirty`, but the relevant section of the source
(https://github.com/obgm/libcoap/blob/develop/src/coap_net.c#L3438) seems to be identical to development tip.

## Problem Description

Resource observation subscription over unreliable transport protocol is sometimes dropped while the session is OK.

### Expected Behavior

https://datatracker.ietf.org/doc/html/rfc7252#section-4.5:
```The recipient SHOULD acknowledge each duplicate copy of a Confirmable message using the same Acknowledgement or Reset message but SHOULD process any request or response in the message only once.```

### Actual Behavior

A duplicate message from a server is correctly processed by the client only once, but wrongly only generates an ACK/RST once, leading the server to remove the observing client from the resource observers list if it does not receive that single ACK (which may occur due to network issues).

### Steps to reproduce

Suppose a client C subscribes to observe a resource on a server S using an underlying unreliable protocol (DTLS in my case).
Every six (6) resource updates (by the way, the value of 6 seems to be hardcoded? Couldn't find a way to configure it),
S makes its message confirmable, expecting an ACK in response. Suppose S sends such a message M with id MID,
it is processed by C, and an ACK is sent by C. Suppose further that the ACK to M is not delivered to S (e.g. network loss).
This makes S retransmit M (with the same MID).
Now consider the code around https://github.com/obgm/libcoap/blob/develop/src/coap_net.c#L3438
The first time M arrives, the check `rcvd->mid == session->last_con_mid` fails, and `last_con_mid` is set to MID in line 3446.
The next time M arrives, the above check passes, and M is silently dropped without being acknowledged.
This leads S to retransmit M until the max threshold (4 by default), and then give up and drop the subscription.

### Code to reproduce this issue

Since network loss of an ACK packet is not easy to reproduce, special code may need to be put together in order to
circumvent sending an ACK by C at some point, and then observe system behaviour. Though code inspection seems sufficient to observe the bug.

## Debug Logs

[...]

PR #1307, opened 2024-01-03
Client fails to ACK a retransmitted CON message

pptz:
This fixes the issue of a client not ACKing a duplicate CON message from the server.
A duplicate CON message is sent by the server if the client's ACK to the original confirmable message did not reach it.
See issue #1306 for full description.

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -3434,7 +3434,11 @@ handle_response(coap_context_t *context, coap_session_t *session,
   if (COAP_PROTO_NOT_RELIABLE(session->proto)) {
     if (rcvd->type == COAP_MESSAGE_CON) {
       if (rcvd->mid == session->last_con_mid) {
-        /* Duplicate response */
+        /* Duplicate response: send ACK/RST, but don't process */
+        if (session->last_con_handler_res == COAP_RESPONSE_OK)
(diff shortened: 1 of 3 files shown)


src/coap_net.c
```diff
@@ -3512,8 +3516,10 @@ handle_response(coap_context_t *context, coap_session_t *session,
                                                      rcvd->mid));
     if (ret == COAP_RESPONSE_FAIL && rcvd->type != COAP_MESSAGE_ACK) {
       coap_send_rst(session, rcvd);
+      session->last_con_handler_res = COAP_RESPONSE_FAIL;
     } else {
       coap_send_ack(session, rcvd);
+      session->last_con_handler_res = COAP_RESPONSE_OK;
     }
   } else {
     coap_send_ack(session, rcvd);
```
mrdeep1, line 3525: You need to add in after this line
````
last_con_handler_res = COAP_RESPONSE_OK;
````
as it is possible that context->response_handler (new line 1512) is now not set, but previously could have been set (but very unlikely).

mrdeep1:
Thanks for doing the update.  Can you please squash the 2 commits into 1.

mrdeep1:
Thanks for doing this PR.

PR #1335, opened 2024-03-03
ESP32: Add platform support

nummada:
This allows the configuration of Libcoap for Expressif ESP32.

LE: I didn't do an exhaustive check related to the full support for adding a platform, but this helped me to configure Libcoap for ESP32.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1120,6 +1120,11 @@ case $host in
     LIBS="${LIBS} -lws2_32"
     ;;
 
+    *xtensa-esp32-elf*)
+    AC_MSG_RESULT([XtensaEsp32Elf])
+    ADDITIONAL_CFLAGS="-D_GNU_SOURCE -DWITH_POSIX"
+    ;;
+
     *)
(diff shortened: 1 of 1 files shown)


mrdeep1:
@nummada I assume that you are aware of https://github.com/espressif/idf-extra-components where the coap directory is the libcoap support (4.3.4) for esp-idf.

nummada:
@mrdeep1 thanks, I was not aware of this.

But I was hoping I could use this: https://github.com/apache/nuttx-apps/tree/master/netutils/libcoap with the esp32 platform.
It's the libcoap support that was recently added into the Nuttx application directory.

mrdeep1:
> LE: I didn't do an exhaustive check related to the full support for adding a platform, but this helped me to
> configure Libcoap for ESP32.

It is possible that there may be some libcoap options / TLS library types that may not get properly supported, but I would expect compile time errors to throw that up.

> But I was hoping I could use this: https://github.com/apache/nuttx-apps/tree/master/netutils/libcoap

Good luck!

PR #1370, opened 2024-05-02
Contiki-NG: Fix compilation error "unknown type name 'fd_set'"

kkrentz:
This resolves a compilation error I encountered with OpenMote targets.


mrdeep1:
Looks fine to me.

PR #1391, opened 2024-05-14
Fixed missing event for server session removal

mopsiok:
There was one place in the code, where server session was closing, but no event about that was generated. When any dynamically allocated app_data was attached to that session, it couldn't be freed properly, causing memory leaks. This PR fixes that.


mrdeep1:
Thanks for finding, fixing and reporting this.

PR #1399, opened 2024-05-17
Fix to cmake build for wolfssl dtls backend

fj-blanco:
The change to `cmake_coap_config.h.in` solves the
following issue: when building with cmake the
functions at `coap_notls.c` were being compiled.


CMakeLists.txt
```diff
@@ -561,6 +561,7 @@ if(ENABLE_DTLS)
     set(WITH_TINYDTLS OFF)
     set(WITH_MBEDTLS OFF)
     set(WITH_OPENSSL OFF)
+    set(WITH_WOLFSSL OFF)
```
mrdeep1, line 564: This change needs to be made to CMakeLists.txt.in as well.

mrdeep1 (changes requested):

CMakeLists.txt
mrdeep1: If you could squash this into one commit, that would be great.

mrdeep1 (approved):

mrdeep1:
Thanks for finding this and creating a fix.

PR #1400, opened 2024-05-17
Contiki-NG: Fix compilation error "implicit declaration of strncasecmp"

kkrentz:
My compiler complained about the implicit declaration of `strncasecmp` in `coap_dtls.c`. I did not pinpoint the commit that introduced this problem. Contiki-NG did not include `strings.h` anyway.


mrdeep1:
Thanks for fixing this.

PR #1424, opened 2024-05-28
coap_net: Fix preprocessor guard

kkrentz:
A preprocessor guard in `coap_net.c` does not take effect since it is misspelled.


mrdeep1:
Thanks for catching this.

PR #1455, opened 2024-07-01
configure.ac: use MbedTLS pkg-config if available

pulsastrix:
Newer MbedTLS versions provide pkg-config files for easier dependency management/linking, see the release notes of versions [3.6.0](https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.6.0) and [2.28.8](https://github.com/Mbed-TLS/mbedtls/releases/tag/v2.28.8).

This PR allows the autoconf build script to utilize the mbedtls pkg-config file if it exists, falling back to the previous method of finding MbedTLS if no pkg-config file was found.

Additionally, this change allows overriding `MbedTLS_CFLAGS` and `MbedTLS_LIBS`, which was previously not possible (unlike all the other DTLS libraries).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -497,19 +497,78 @@ if test "x$build_dtls" = "xyes"; then
                   [have_wolfssl="yes"],
                   [have_wolfssl="no"])
 
-    # Mbed TLS [does not have mbedtls.pc pkg-config file]
-    AC_CHECK_LIB(mbedtls, mbedtls_version_get_string,
-                 [have_mbedtls="yes"; MbedTLS_CFLAGS="" ; MbedTLS_LIBS="-lmbedtls -lmbedcrypto -lmbedx509"],
-                 [have_mbedtls="no"], -lmbedx509 -lmbedcrypto)
+    # Mbed TLS
+    if test "x${MbedTLS_CFLAGS+set}" = "xset"; then
+        mbedtls_cflags_overridden="yes"
+    else
+        mbedtls_cflags_overridden="no"
+    fi
+
+    if test "x${MbedTLS_LIBS+set}" = "xset"; then
+        mbedtls_libs_overridden="yes"
+    else
+        mbedtls_libs_overridden="no"
+    fi
+
+    # If MbedTLS_CFLAGS and MbedTLS_LIBS are overridden, pkg-config will always assume the library was found, even
+    # though MbedTLS doesn't necessarily have a pkg-config file. Therefore, we might as well use the "old" way to
+    # determine whether we can link MbedTLS (and which version we have).
+    if test "x$mbedtls_libs_overridden" = "xyes" -a "x$mbedtls_cflags_overridden" = "xyes"; then
+      have_mbedtls="no"
+      mbedtls_has_pkgconfig="no"
+    else
+      # Attempt to find MbedTLS using pkg-config.
+
+      # When statically linking against libcoap, all transitive dependencies need to be specified as linker flags
+      # as well. Use PKG_CHECK_MODULES_STATIC for that.
+      if test "x$enable_static" = "xyes"; then
+        PKG_CHECK_MODULES_STATIC([MbedTLS],
+                                 [mbedtls],
+                                 [have_mbedtls="yes"; mbedtls_has_pkgconfig="yes"],
+                                 [have_mbedtls="no"; mbedtls_has_pkgconfig="no"])
+      else
+        PKG_CHECK_MODULES([MbedTLS],
+                          [mbedtls],
+                          [have_mbedtls="yes"; mbedtls_has_pkgconfig="yes"],
+                          [have_mbedtls="no"; mbedtls_has_pkgconfig="no"])
+      fi
+    fi
+
+    if test "x$have_mbedtls" = "xno"; then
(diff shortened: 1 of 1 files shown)


mrdeep1:
Thanks for doing this.

Great that MBed TLS have eventually added .pc files.  Shame that they do not have it for their make system, only for cmake.

However, for libcoap
````
./configure --with-mbedtls --disable-doxygen --disable-manpages --enable-tests --disable-shared
make
````
fails with 
````
make[2]: Entering directory '/home/jon/libcoap/tests'
  CCLD     testdriver
/usr/bin/ld: ../.libs/libcoap-3-mbedtls.a(libcoap_3-mbedtls_la-coap_mbedtls.o): undefined reference to symbol 'mbedtls_x509_crt_parse_file'
/usr/bin/ld: /usr/local/lib/libmbedx509.so.7: error adding symbols: DSO missing from command line
````
or
````
./configure --with-mbedtls --disable-doxygen --disable-manpages --disable-tests --disable-shared
make
````
gives
````
  CCLD     coap-client-mbedtls
/usr/bin/ld: ../.libs/libcoap-3-mbedtls.a(libcoap_3-mbedtls_la-coap_mbedtls.o): undefined reference to symbol 'mbedtls_x509_crt_parse_file'
/usr/bin/ld: /usr/local/lib/libmbedx509.so.7: error adding symbols: DSO missing from command line
````
The issue here is the non-shared file `../.libs/libcoap-3-mbedtls.a` does not indicate that the libraries mbedcrypto mbedx509 are not indicated as being needed.
````
./configure --with-mbedtls --disable-doxygen --disable-manpages --disable-tests --enabled-shared
make
````
does work as expected (note: `tests/testdriver` has to use the static file, hence why that is always failing).

I suspect that you may need to use `PKG_CHECK_MODULES_STATIC` if `--disable-shared` is set.

**Comment by pulsastrix** [comment]
Hi, thanks for the review.

I have now replaced `PKG_CHECK_MODULES` with  `PKG_CHECK_MODULES_STATIC` if `--enable-static` is set, which should fix the issue you mentioned (at least it did on my system).

File: configure.ac [hunk 1]
```diff
@@ -497,19 +497,78 @@ if test "x$build_dtls" = "xyes"; then
...
-    AC_CHECK_LIB(mbedtls, mbedtls_version_get_string,
-                 [have_mbedtls="yes"; MbedTLS_CFLAGS="" ; MbedTLS_LIBS="-lmbedtls -lmbedcrypto -lmbedx509"],
-                 [have_mbedtls="no"], -lmbedx509 -lmbedcrypto)
+    # Mbed TLS
+    if test "x${MbedTLS_CFLAGS+set}" = "xset"; then
+        mbedtls_cflags_overridden="yes"
+    else
+        mbedtls_cflags_overridden="no"
+    fi
+    # Mbed TLS
```
mrdeep1, line 506: `Mbed TLS` is unnecessarily duplicated

configure.ac
```diff
@@ -497,19 +497,78 @@ if test "x$build_dtls" = "xyes"; then
...
+    if test "x${MbedTLS_LIBS+set}" = "xset"; then
+        mbedtls_libs_overridden="yes"
+    else
+        mbedtls_libs_overridden="no"
+    fi
+
+    # If MbedTLS_CFLAGS and MbedTLS_LIBS are overridden, pkg-config will always assume the library was found, even
+    # though MbedTLS doesn't necessarily have a pkg-config file. Therefore, we might as well use the "old" way to
+    # determine whether we can link MbedTLS (and which version we have).
+    if test "x$mbedtls_libs_overridden" = "xyes" -a "x$mbedtls_cflags_overridden" = "xyes"; then
```
mrdeep1, line 516: I think that this should be (OR)
````
  if test "x$mbedtls_libs_overridden" = "xyes" -o "x$mbedtls_cflags_overridden" = "xyes"; then
````

mrdeep1 (changes requested):
@pulsastrix It is looking good.  Just a couple of things.

configure.ac
pulsastrix: It was actually intended to be AND.

The issue I was trying to avoid here is the following scenario:
- Both `MbedTLS_CFLAGS` and `MbedTLS_LIBS` are set as environment variables/overridden.
- The MbedTLS version provided through these environment variables does not have a pkg-config file.

In this situation, `PKG_CHECK_MODULES` would simply assume that the library was found without ever executing pkg-config, at least based on my observations. [^1]

Later on, the configure-script would then want to check the version using pkg-config (line 559), and fail because there actually is no pkg-config file to get the version from. 

If either `MbedTLS_CFLAGS` or `MbedTLS_LIBS` are overridden (but not both) this issue cannot occur, as `PKG_CHECK_MODULES` will run pkg-config to determine the missing variable.

[^1]: This behavior is at least hinted at in the [man pages](https://manpages.debian.org/testing/pkg-config/pkg-config.1.en.html#AUTOCONF_MACROS), although it is not outright stated.

mrdeep1: Testing on Ubuntu 22 confirms this needs to be an AND here.

mrdeep1 (approved):

PR #1464, opened 2024-07-11
enable HAVE_STRUCT_CMSGHDR when building with mingw

anyc:
When building libcoap with mingw, I noticed that coap_session_get_ifindex() always return zero. In the code, it looks like this is always set to zero if HAVE_STRUCT_CMSGHDR is unset in line: https://github.com/obgm/libcoap/blob/7d5e5f3b8c31f7af60b697659f9a97de7db9a3cc/src/coap_io.c#L1165

Wouldn't -1 be a better value to indicate an undefined value?

If libcoap is built for Windows without mingw, HAVE_STRUCT_CMSGHDR is always defined. In this PR, I removed the special handling of mingw, added a few aliases and the code seems to work with HAVE_STRUCT_CMSGHDR defined and mingw.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,7 +300,7 @@ check_function_exists(random HAVE_RANDOM)
 check_function_exists(if_nametoindex HAVE_IF_NAMETOINDEX)
 
 # check for symbols
-if(WIN32 AND NOT MINGW)
+if(WIN32)
   set(HAVE_STRUCT_CMSGHDR 1)
   message(STATUS "setting HAVE_STRUCT_CMSGHDR")
 elseif(${CMAKE_SYSTEM_NAME} STREQUAL QNX)
--- a/src/coap_io.c
+++ b/src/coap_io.c
@@ -86,6 +86,16 @@ coap_mfree_endpoint(coap_endpoint_t *ep) {
 }
 #endif /* COAP_SERVER_SUPPORT */
 
+#if defined(__MINGW32__)
+#if(_WIN32_WINNT >= 0x0600)
+#define CMSG_FIRSTHDR WSA_CMSG_FIRSTHDR
+#define CMSG_NXTHDR WSA_CMSG_NXTHDR
+#define CMSG_LEN WSA_CMSG_LEN
+#define CMSG_SPACE WSA_CMSG_SPACE
+#define cmsghdr _WSACMSGHDR
+#endif /* (_WIN32_WINNT>=0x0600) */
+#endif /* defined(__MINGW32__) */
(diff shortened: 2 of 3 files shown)


src/coap_io.c
```diff
@@ -86,6 +86,16 @@ coap_mfree_endpoint(coap_endpoint_t *ep) {
...
 #endif /* COAP_SERVER_SUPPORT */
 
+#if defined(__MINGW32__)
+#if(_WIN32_WINNT >= 0x0600)
+#define CMSG_FIRSTHDR WSA_CMSG_FIRSTHDR
+#define CMSG_NXTHDR WSA_CMSG_NXTHDR
+#define CMSG_LEN WSA_CMSG_LEN
+#define CMSG_SPACE WSA_CMSG_SPACE
+#define cmsghdr _WSACMSGHDR
+#endif //(_WIN32_WINNT>=0x0600)
```
mrdeep1, line 96: Coding style is to use `/* ... */`, not `//`

CMakeLists.txt
```diff
@@ -300,7 +300,7 @@ check_function_exists(random HAVE_RANDOM)
 check_function_exists(if_nametoindex HAVE_IF_NAMETOINDEX)
 
 # check for symbols
-if(WIN32 AND NOT MINGW)
+if(WIN32)
```
mrdeep1, line 303: This change needs to be in `CMakeLists.txt.in` as well.

mrdeep1 (commented):
Couple of minor things.

src/coap_io.c
```diff
@@ -734,9 +744,11 @@ struct in_pktinfo {
 
 #if defined(_WIN32)
 #include <mswsock.h>
-#if !defined(__MINGW32__)
+#if defined(__MINGW32__)
+static __thread LPFN_WSARECVMSG lpWSARecvMsg = NULL;
+#else
```
mrdeep1, line 749: I would prefer this to be
````
#else /* ! __MINGW32__ */
````
as what follows the else is not mingw.

File: src/coap_io.c [hunk 4]
```diff
@@ -734,9 +744,11 @@ struct in_pktinfo {
...
 static __declspec(thread) LPFN_WSARECVMSG lpWSARecvMsg = NULL;
-#endif /* ! __MINGW32__ */
+#endif /* defined(__MINGW32__) */
```
mrdeep1, line 751: This should then be reverted, so indicating the end of the code block to be what it is finishing.

mrdeep1 (commented):
Minor documentation fixes

PR #1495, opened 2024-08-12
coap_proxy.c: Fix error "'.*' expects argument of type 'int'"

kkrentz:
Adds a cast to int in order to avoid a compilation error on an embedded target.


mrdeep1:
Thinks for fixing this.

PR #1496, opened 2024-08-12
Contiki-NG: Fix coap_resolve_address_info

kkrentz:
`addr_ip` was declared as a pointer, but sometimes not used in this manner.

### DIFF
--- a/src/coap_address.c
+++ b/src/coap_address.c
@@ -729,23 +729,17 @@ coap_resolve_address_info(const coap_str_const_t *address,
 #elif defined(WITH_CONTIKI)
 
 #include <os/net/ipv6/uiplib.h>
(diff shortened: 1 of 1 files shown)


mrdeep1:
Thanks for this fix.

It would be great if `coap_resolve_address_info()` could also resolve DNS provided names, but I have no idea as to how to do that in Contiki-NG.

kkrentz:
Contiki-NG provides the `resolv` module. However, it may return `RESOLV_STATUS_RESOLVING` and provide the result later. This makes it difficult to use it here. Another complication is that modules are not usually compiled and only added on demand.

mrdeep1:
Thanks for the feedback re `resolv`.  I'll ignore doing this for now.

PR #1505, opened 2024-08-26
coap_net.h: add ifdef WITH_LWIP around lwip functions

anyc:
Hello,

it looks like it is easily possible to generate different language bindings for libcoap with swig. I only had to make the changes in this PR to generate bindings for Python and (after a small patch in the generated code) for C#.

Would you be interested in a PR that adds some documentation on how to use swig to generate the bindings for the languages above?

Thank you!


anyc:
Hm, I don't know what to do about this failed check. Would it be necessary to process coap_lwip.txt.in only if WITH_LWIP is set?

mrdeep1:
My suggestion would be
````
diff --git a/include/coap3/coap_net.h b/include/coap3/coap_net.h
index 6e0afde8..455bab3b 100644
--- a/include/coap3/coap_net.h
+++ b/include/coap3/coap_net.h
@@ -803,6 +803,7 @@ COAP_API void coap_io_do_epoll(coap_context_t *ctx, struct epoll_event *events,

 /**@}*/

+#if defined(WITH_LWIP) || defined(WITH_LWIP_MAN_CHECK)
 /**
  * @ingroup application_api
  * @defgroup lwip LwIP specific API
@@ -848,6 +849,7 @@ void coap_lwip_set_input_wait_handler(coap_context_t *context,
                                       void *input_arg);

 /**@}*/
+#endif /* WITH_LWIP || WITH_LWIP_MAN_CHECK */

 /**
  * @deprecated Use coap_io_process() instead.
diff --git a/man/examples-code-check.c b/man/examples-code-check.c
index bb1f4691..623ed235 100644
--- a/man/examples-code-check.c
+++ b/man/examples-code-check.c
@@ -168,6 +168,9 @@ check_synopsis(const char *file) {
     exit_code = 1;
     return;
   }
+  if (!strcmp(file, "coap_lwip.txt.in")) {
+    fprintf(fpcode, "#define WITH_LWIP_MAN_CHECK\n");
+  }
   fprintf(fpcode, "#include <coap3/coap.h>\n");
   fprintf(fpcode, "#ifdef __GNUC__\n");
   fprintf(fpcode, "#define U __attribute__ ((unused))\n");
````

mrdeep1:
> Would you be interested in a PR that adds some documentation on how to use swig to generate the bindings
> for the languages above?

Makes sense.  Perhaps the WiKi is a good place.

anyc:
We also need an additional file as input for swig, the C# bindings need a patch and maybe it would be good to add a little example for each binding. Perhaps it would be good to also add this to the git repository - maybe under doc/ or examples/ ?

mrdeep1 (commented):
Need to let Doxygen document these LwIP functions.

mrdeep1:
> Perhaps it would be good to also add this to the git repository - maybe under doc/ or examples/ ?

Ok - suggest under `examples/swig`.  A README and example files.

PR #1521, opened 2024-09-17
Improve MTU Handling

kkrentz:
The main issue that this PR solves is that on Contiki-NG one can create PDUs that will not fit in the uIP buffer. The problem is that the value of `COAP_DEFAULT_MAX_PDU_RX_SIZE` is incorrect and a check in `coap_pdu_init` is also flawed. On LWIP, there seems to be an analogous problem.


include/coap3/coap_pdu_internal.h
```diff
@@ -75,9 +75,9 @@
 
 #ifndef COAP_DEFAULT_MAX_PDU_RX_SIZE
 #if defined(WITH_LWIP)
-#define COAP_DEFAULT_MAX_PDU_RX_SIZE (COAP_MAX_MESSAGE_SIZE_TCP16+4UL)
+#define COAP_DEFAULT_MAX_PDU_RX_SIZE (COAP_MAX_MESSAGE_SIZE_TCP16)
```
mrdeep1: I think the original is still correct - it defines the maximum space for TCP (16) data (including tokens, options etc) within the PDU as well as the CoAP header being added on.

mrdeep1 (commented):
Thanks for this.  Other than I think the LwIP original code is correct, the commits need to be squashed into a single commit.

include/coap3/coap_pdu_internal.h
kkrentz: I see. I got confused by the `assert(size <= COAP_DEFAULT_MAX_PDU_RX_SIZE);` in `pdu_init`. So, only `pdu_init` needs fixing.

mrdeep1: Yes, I had to think things carefully through!

PR #1530, opened 2024-10-10
coap_join_mcast_group_intf: fix interface selection on windows

anyc:
On the Windows platform, a set `ifname` parameter of the `coap_join_mcast_group_intf` function is only used for debug output. Only if it is `NULL`, the interface which should join a multicast group is derived from the bind address. So, setting the `ifname` parameter on the windows platform is counterproductive.

With this patch, `ifname` is always set to `NULL` on the windows platform and libcoap can try to derive the interface from the bind address.

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -4706,7 +4706,14 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
       }
     }
   }
-#endif /* ! _WIN32 */
+#else /* _WIN32 */
+  /*
+   * On Windows this function ignores the ifname variable so we unset this
+   * variable on this platform in any case in order to enable the interface
+   * selection from the bind address below.
+   */
+  ifname = 0;
+#endif /* _WIN32 */
 
   /* Add in mcast address(es) to appropriate interface */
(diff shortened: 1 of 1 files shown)


src/coap_net.c
```diff
@@ -4706,6 +4706,13 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
       }
     }
   }
+#else /* ! _WIN32 */
```
mrdeep1: We are entering a section where _WIN32 is defined, so `!` is confusing

src/coap_net.c
```diff
@@ -4706,6 +4706,13 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
...
+  /*
+   * On Windows this function ignores the ifname variable so we unset this
+   * variable on this platform in any case in order to enable the interface
+   * selection from the bind address below.
+   */
+  ifname = 0;
 #endif /* ! _WIN32 */
```
mrdeep1: Likewise, this is the end of the _WIN32 section, so `!` should be removed.

mrdeep1 (commented):
Minor documentation tidy up - otherwise this looks good to me.

mrdeep1:
Thanks for doing this.

PR #1535, opened 2024-10-21
Contiki: Post a dedicated event when scheduling an immediate prepare_io

kkrentz:
I realized that, e.g., button drivers, broadcast a `PROCESS_EVENT_POLL`. This causes all processes to wake up, including our `libcoap_io_process`. Moreover, enclosed data will be treated as a coap_context_t *. This PR fixes this issue by allocating a dedicated event.

### DIFF
--- a/src/coap_io_contiki.c
+++ b/src/coap_io_contiki.c
@@ -17,11 +17,15 @@
 #include "coap3/coap_libcoap_build.h"
 #include "contiki-net.h"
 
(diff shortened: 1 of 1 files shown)


mrdeep1:
In principle looks fine to me without doing any testing.  Frustrating that it is not possible to release the dedicated event.

kkrentz:
My bad. In the official code base nobody is broadcasting a `PROCESS_EVENT_POLL` and it would probably break many other things, too.

PR #1558, opened 2024-12-16
doc: fix invisible text if browser is set to prefer dark mode

pulsastrix:
Recent builds of the libcoap documentation use a Doxygen version that supports and automatically enables dark mode if a preference for dark mode has been indicated by the user.

Unfortunately, this caused some issues with the CSS overrides for text colors set in `doc/docbook.local.css`, which make some text invisible or hard to read in dark mode.

This PR fixes the issue by setting different (more readable) text colors if in dark mode (while leaving the text colors for light mode untouched).

For comparison:

Current build on `develop` branch:
image

With the changes made in this PR:
image

### DIFF
--- a/doc/docbook.local.css
+++ b/doc/docbook.local.css
@@ -45,3 +45,26 @@ a.st-synopsis:visited {
 a.st-synopsis:hover {
 	text-decoration: underline;
 }
+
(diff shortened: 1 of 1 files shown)


mrdeep1:
Thanks for fixing this - looks good to me.

PR #1560, opened 2024-12-19
coap_subscribe_internal.h: Fix "unbalanced grouping commands"

kkrentz:
This fixes that the CI task `documentation` sometimes failed with "unbalanced grouping commands".


mrdeep1:
Good catch and thanks.

PR #1581, opened 2025-02-13
Contiki-NG: Fix "unknown type name 'u_char'"

kkrentz:
When compiling for an embedded target, Contiki-NG fails with `unknown type name 'u_char'`. This PR fixes this problem.


mrdeep1:

PR #1590, opened 2025-02-20
coap_threadsafe_internal.h: Fix coap_lock_invert.

peterharperuk:
This code fails to compile if COAP_THREAD_SAFE is false as the macro parameter is "alt_lock" not "func".


mrdeep1:

PR #1599, opened 2025-03-11
coap_io: do not call select() for contexts without sockets on Windows

mkicherer:
Hello,

in a NodeJS environment, my wrapper library for libcoap maintains the list of created contexts and provides a loop() function that will call `coap_io_process` with every context and `COAP_IO_NO_WAIT` every few milliseconds (as we do not have threads or proper access to the internal event loop). If I create an "empty" context without sessions or endpoints early on while this loop is running, we see errors from the `select()` call in coap_io.c if we start this application on Windows. In this application, we start client sessions in this context later on certain events.

After comparing the documentation for `select()` on Windows [1] and Linux [2], it looks to me like the Windows version does not like being called without file descriptors/sockets. I applied this patch and it seems to work without error messages so far.

I could add code in my wrapper to call `coap_io_process` only for contexts with active sockets but I think this patch might also benefit others. Maybe it would even make sense to skip more code in this function if the context has no sockets. What do you think?

Thank you!

[1] https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select
[2] https://man7.org/linux/man-pages/man2/select.2.html

### DIFF
--- a/src/coap_io.c
+++ b/src/coap_io.c
@@ -1848,13 +1848,18 @@ coap_io_process_with_fds_lkd(coap_context_t *ctx, uint32_t timeout_ms,
     tv.tv_sec = (long)(timeout / 1000);
   }
 
-  /* Unlock so that other threads can lock/update ctx */
-  coap_lock_unlock(ctx);
+  /* on Windows select will return an error if called without FDs */
+  if (nfds > 0) {
+    /* Unlock so that other threads can lock/update ctx */
+    coap_lock_unlock(ctx);
 
-  result = select((int)nfds, &ctx->readfds, &ctx->writefds, &ctx->exceptfds,
-                  timeout > 0 ? &tv : NULL);
+    result = select((int)nfds, &ctx->readfds, &ctx->writefds, &ctx->exceptfds,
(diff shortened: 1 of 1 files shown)


src/coap_io.c
```diff
@@ -1848,13 +1848,18 @@ coap_io_process_with_fds_lkd(coap_context_t *ctx, uint32_t timeout_ms,
     tv.tv_sec = (long)(timeout / 1000);
   }
 
-  /* Unlock so that other threads can lock/update ctx */
-  coap_lock_unlock(ctx);
+  // on Windows select will return an error if called without FDs
```
mrdeep1: libcoap's code comment style is `/* ..... */`, not `// ....`

Please update this accordingly.

mrdeep1:
It is also true that even if there are endpoints / sessions associated with a context, it is conceivable (but very unlikely) that there is nothing that is in a read /write / except state.

I'm not so sure about skipping more code as there are some subsequent timeout checks, and the time spent in the function should always be reported (even if it is 0 milli-secs).

PR #1622, opened 2025-03-26
coap-client: let observe duration override the default wait timeout

mkicherer:
It may be a bit surprising when observe durations greater than 90 seconds seem to be ignored.


mrdeep1:

PR #1672, opened 2025-05-22
OSCORE: Validate a non-OSCORE response to an OSCORE request

pawszw2:
According to RFC 8613, section 2, a response to an OSCORE request SHALL contain the OSCORE option. According to this section and section 8.3, the only exceptions to this rule are OSCORE erros, described in steps 2, 3, and 6 of section 8.2: Bad Option, Unauthorized and Bad Request. In all other cases the response violates RFC.

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -4305,6 +4305,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
   coap_pdu_t *dec_pdu = NULL;
 #endif /* COAP_OSCORE_SUPPORT */
   int is_ext_token_rst;
+  int oscore_invalid = 0;
 
   pdu->session = session;
   coap_show_pdu(COAP_LOG_DEBUG, pdu);
@@ -4435,6 +4436,14 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
       coap_log_debug("Decrypted PDU\n");
       coap_show_pdu(COAP_LOG_DEBUG, pdu);
     }
+  } else if (COAP_PDU_IS_RESPONSE(pdu) &&
+             session->oscore_encryption &&
+             pdu->type != COAP_MESSAGE_RST) {
+    if (COAP_RESPONSE_CLASS(pdu->code) == 2) {
+      /* Violates RFC 8613 2 */
+      coap_log_err("received an invalid response to the OSCORE request\n");
+      oscore_invalid = 1;
+    }
   }
 #endif /* COAP_OSCORE_SUPPORT */
 
@@ -4449,7 +4458,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
         /* Flush out any entries on session->delayqueue */
         coap_session_connected(session);
     }
-    if (coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+    if (oscore_invalid || coap_option_check_critical(session, pdu, &opt_filter) == 0) {
       packet_is_bad = 1;
       goto cleanup;
     }
@@ -4623,8 +4632,8 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
     goto cleanup;
 
   case COAP_MESSAGE_NON:
-    /* check for unknown critical options */
-    if (coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+    /* check for oscore issue or unknown critical options */
+    if (oscore_invalid || coap_option_check_critical(session, pdu, &opt_filter) == 0) {
       packet_is_bad = 1;
       coap_send_rst_lkd(session, pdu);
       goto cleanup;
@@ -4634,7 +4643,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
     }
     break;
 
-  case COAP_MESSAGE_CON:        /* check for unknown critical options */
+  case COAP_MESSAGE_CON:
     /* In a lossy context, the ACK of a separate response may have
      * been lost, so we need to stop retransmitting requests with the
      * same token. Matching on token potentially containing ext length bytes.
@@ -4642,8 +4651,9 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
     /* find message token in sendqueue to stop retransmission */
     coap_remove_from_queue_token(&context->sendqueue, session, &pdu->actual_token, &sent);
 
-    if (!COAP_PDU_IS_SIGNALING(pdu) &&
-        coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+    /* check for oscore issue or unknown critical options in non-signaling messages */
+    if (oscore_invalid ||
+        (!COAP_PDU_IS_SIGNALING(pdu) && coap_option_check_critical(session, pdu, &opt_filter) == 0)) {
       packet_is_bad = 1;
       if (COAP_PDU_IS_REQUEST(pdu)) {
         response =


mrdeep1:
Thanks for raising this.

When I read [RFC8613 2. The OSCORE Option](https://datatracker.ietf.org/doc/html/rfc8613#section-2), I read (some time back)
````
A successful response to a request with the OSCORE option SHALL 
contain the OSCORE option.
````
as the generator of the the response must include the OSCORE option.  I did not read it from the perspective of the recipient and what the recipient is supposed to do if there is not a OSCORE option.

If the recipient of a request that contains the OSCORE option, and does not know what to do with it, as it is a critical option, it falls under [RFC 7252 5.4.1. Critical/Elective](https://datatracker.ietf.org/doc/html/rfc7252#section-5.4.1) and so must generate an appropriate rejection response (which is unlikely to contain an OSCORE option).

Looking at the code, your code addition (which has compilation errors)  sits between
````
Line 4381 src/coap_net.c
if (coap_check_option(pdu, COAP_OPTION_OSCORE, &opt_iter) != NULL) {
      ...
     <Your additional code>
Old line 4438, your line 4469 src/coap_net.c
}
````
so is only potentially used when the OSCORE option actually exists in the incoming PDU. So I am not sure what you are trying to fix here.

**Comment by pawszw2** [comment]
Hi @mrdeep1 , thanks for a quick reply. Actually my code is meant to address the case in which a Client sends an OSCORE request, but receives a message without OSCORE option as a response. The current libcoap behavior is to accept this message without any validation, which may be a security concern. According to mentioned in PR description sections of RFC, there are only 3 cases in which receiving a non-OSCORE response to an OSCORE request is valid, and they are due to an error in processing OSCORE request on the server side. Receiving a successful response without OSCORE option could mean that some adversary is trying to impersonate a trusted side.
Introduced in PR "else if" will be checked when there is no OSCORE option in the response. Git diff suggests that it is inside previous "if" statement, but it actually isn't.
Sorry for compilation errors, I had to check the solution on the old libcoap version and made a mistake when copy pasting. I will fix remaining errors soon.
Feel free to ask further questions.

File: src/coap_net.c [hunk 1]
```diff
@@ -4435,6 +4435,42 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
       coap_log_debug("Decrypted PDU\n");
       coap_show_pdu(COAP_LOG_DEBUG, pdu);
     }
+  } else if (COAP_PDU_IS_RESPONSE(pdu) &&
+             session->oscore_encryption &&
+             pdu->type != COAP_MESSAGE_RST) {
+    if (COAP_RESPONSE_CODE_BAD_OPTION != pdu->code &&
+        COAP_RESPONSE_CODE_UNAUTHORIZED != pdu->code &&
+        COAP_RESPONSE_CODE_BAD_REQUEST != pdu->code) {
```
mrdeep1: As per https://github.com/obgm/libcoap/pull/571#issuecomment-727811702 and https://github.com/obgm/libcoap/pull/571#issuecomment-728191913 in #571 this should read
````
   if ( COAP_RESPONSE_CODE(402) != pdu->code &&
        COAP_RESPONSE_CODE(401) != pdu->code &&
        COAP_RESPONSE_CODE(400) != pdu->code) {
````
or preferably as the way the rest of the coding style does it
````
    if (pdu->code != COAP_RESPONSE_CODE(402) &&
        pdu->code != COAP_RESPONSE_CODE(401) &&
        pdu->code != COAP_RESPONSE_CODE(400) {
````
or even better
````
    if (COAP_RESPONSE_CLASS(response->code) == 2) {
````
as 4.xx and 5.xx can be unprotected.  See [RFC 8613 B.2.1. Implementation Considerations](https://datatracker.ietf.org/doc/html/rfc8613#appendix-B.2.1).  This should be OK for the first CSM [RFC 8163 4.3. Signalling Messages](https://datatracker.ietf.org/doc/html/rfc8613#section-4.3), but needs testing.

File: src/coap_net.c [hunk 2]
```diff
@@ -4435,6 +4435,42 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
...
+        /* find message id in sendqueue to stop retransmission */
+        coap_remove_from_queue(&context->sendqueue, session, pdu->mid, &sent);
+        if (sent && session->con_active) {
+          session->con_active--;
+          if (session->state == COAP_SESSION_STATE_ESTABLISHED)
+            /* Flush out any entries on session->delayqueue */
+            coap_session_connected(session);
+        }
+        goto cleanup;
+        break;
```
mrdeep1: No need for` break;`

src/coap_net.c
```diff
@@ -4435,6 +4435,42 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
...
+      case COAP_MESSAGE_CON:
+        /* find message token in sendqueue to stop retransmission */
+        coap_remove_from_queue_token(&context->sendqueue, session, &pdu->actual_token, &sent);
+      /* Intentionally without break statement */
```
mrdeep1: This needs to be
````
        /* Fall through */
````

File: src/coap_net.c [hunk 4]
```diff
@@ -4435,6 +4435,42 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
...
+      case COAP_MESSAGE_NON:
+        coap_send_rst_lkd(session, pdu);
+        goto cleanup;
+        break;
```
mrdeep1: No need for `break;`

src/coap_net.c
```diff
@@ -4435,6 +4435,42 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
...
+      case COAP_MESSAGE_RST:
+      /* added to fix compilation warning */
```
mrdeep1: Comment not really needed - if needed, it should be the line above.

mrdeep1 (changes requested):
Code changes required

src/coap_net.c
pawszw2: Good to see a greater picture. I agree, that response classes 5 and 7 should also be accepted, but I wander if all 4 class response codes should be accepted, or only the mentioned ones. So I am thinking about following change:
 ```suggestion
    if (COAP_RESPONSE_CLASS(pdu->code) < 5 &&
        pdu->code != COAP_RESPONSE_CODE(402) &&
        pdu->code != COAP_RESPONSE_CODE(401) &&
        pdu->code != COAP_RESPONSE_CODE(400) ) {
```

The one you suggested looks better:
```suggestion
    if (COAP_RESPONSE_CLASS(response->code) == 2) {
```
but I am not sure if there are some security considerations that should be taken into account. I have to give it a thought.

mrdeep1:
I am a bit nervous about having a subset of code duplicated on detection of an invalid non-OSCORE response because there may need to be a subsequent change to the original code copied for the duplication and the duplicated code does not get updated.  The duplicated code is also a lot of code that will rarely, if ever, get used.

My suggestion would to go for something like
````
diff --git a/src/coap_net.c b/src/coap_net.c
index cfb6bbb3..297edc97 100644
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -4305,6 +4305,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
   coap_pdu_t *dec_pdu = NULL;
 #endif /* COAP_OSCORE_SUPPORT */
   int is_ext_token_rst;
+  int oscore_invalid = 0;

   pdu->session = session;
   coap_show_pdu(COAP_LOG_DEBUG, pdu);
@@ -4435,6 +4436,17 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
       coap_log_debug("Decrypted PDU\n");
       coap_show_pdu(COAP_LOG_DEBUG, pdu);
     }
+  } else if (COAP_PDU_IS_RESPONSE(pdu) &&
+             session->oscore_encryption &&
+             pdu->type != COAP_MESSAGE_RST) {
+    if (COAP_RESPONSE_CLASS(pdu->code) < 5 &&
+        pdu->code != COAP_RESPONSE_CODE(402) &&
+        pdu->code != COAP_RESPONSE_CODE(401) &&
+        pdu->code != COAP_RESPONSE_CODE(400)) {
+      /* Violates RFC 8613 2 */
+      coap_log_err("received an invalid response to the OSCORE request\n");
+      oscore_invalid = 1;
+    }
   }
 #endif /* COAP_OSCORE_SUPPORT */

@@ -4449,7 +4461,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
         /* Flush out any entries on session->delayqueue */
         coap_session_connected(session);
     }
-    if (coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+    if (oscore_invalid || coap_option_check_critical(session, pdu, &opt_filter) == 0) {
       packet_is_bad = 1;
       goto cleanup;
     }
@@ -4623,8 +4635,8 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
     goto cleanup;

   case COAP_MESSAGE_NON:
-    /* check for unknown critical options */
-    if (coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+    /* check for oscore issue or unknown critical options */
+    if (oscore_invalid || coap_option_check_critical(session, pdu, &opt_filter) == 0) {
       packet_is_bad = 1;
       coap_send_rst_lkd(session, pdu);
       goto cleanup;
@@ -4643,7 +4655,7 @@ coap_dispatch(coap_context_t *context, coap_session_t *session,
     coap_remove_from_queue_token(&context->sendqueue, session, &pdu->actual_token, &sent);

     if (!COAP_PDU_IS_SIGNALING(pdu) &&
-        coap_option_check_critical(session, pdu, &opt_filter) == 0) {
+        (oscore_invalid || coap_option_check_critical(session, pdu, &opt_filter) == 0)) {
       packet_is_bad = 1;
       if (COAP_PDU_IS_REQUEST(pdu)) {
         response =
````
At some point all the commits need to get merged into a single commit  with an appropriate description.

src/coap_net.c
mrdeep1: 4.08 [RFC9177 5. [The Use of the 4.08 (Request Entity Incomplete) Response Code](https://datatracker.ietf.org/doc/html/rfc9177#name-the-use-of-the-408-request-)](https://datatracker.ietf.org/doc/html/rfc9177#section-5) is another one that should not be checked for.  It is likely that there could be others and new 4.xx may get introduced in new RFCs.

So, again, I think the only safe answer here is
````
     if (COAP_RESPONSE_CLASS(response->code) == 2) {
````
Which catches a 'good' response being used maliciously.

pawszw2: I see, let's stay then with the check for response class 2.

pawszw2:
I've addressed your comments, also regarding the commit history. Is there anything else to do?

mrdeep1:
It all looks good to me.  Thanks for doing this.

pawszw2:
Thanks for review and integrating this.

pawszw2:
Hi @mrdeep1, I have a question - is there a plan to make another official libcoap release? I can't find any details on libcoap release strategy.

mrdeep1:
Yes, it is planned that there will be future official libcoap releases.  There are several almost about to be CoAP RFCs being worked on which I would like to see in the next libcoap release, hence timescales for the release are unknown at present.

Issue #1696, opened 2025-06-18
Possibility to trigger async multiple times


mkicherer:
Hello,

this time I only have a question/feature request. In my server, I have resources where their value depends on multiple async function calls. As far as I understood libcoap's async functions, if I call `coap_async_trigger` the async object is deleted automatically after the corresponding request handler is called again. If I need to do multiple async function calls, it would be helpful if I could trigger an async object multiple times and, for example, either delete it manually or automatically only after I set a response code.

I could create my own state machine which only calls `coap_async_trigger` in the final state but one of my use cases is a kind-of `.well-known/core` alternative which also returns the values of the resources. In this case, it would be really helpful if I could trigger an async multiple times as I would not have to differentiate in my code if this is a specific request for this resource (where I can use libcoap's async) or a "bulk" request (where I would need to trigger my own state machine).

As far as I understand libcoap's code so far, it looks comparatively easy to achieve this as we would have to make the `coap_free_async_lkd` call in `coap_check_async` (coap_net.c:4977) conditional and reset the async delay. Maybe I could call `coap_async_set_delay` in the request handler if I want to call `coap_async_trigger` again later and we only call `coap_free_async_lkd` in `coap_check_async` if `(async->delay != 0 && async->delay < now)`?

If you think this is viable I would try to implement this and create a PR.

Thank you!

mrdeep1:
I'm struggling with trying to understand what you are looking to do here.  Is it you need to keep on calling the request handler until the response is in some sort of stable state?

Async support was added in way back to provide support for separate responses [RFC7252 5.2.2 Separate](https://rfc-editor.org/rfc/rfc7252#section-5.2.2) where the request was acknowledged with an empty ACK and later followed by the actual response.  Your Async basic understanding seems to be correct.

Then came along the ability to observe resources [RFC7641](https://datatracker.ietf.org/doc/html/rfc7641) where the notion of unsolicited responses was added in (tracked by a common token).  Are you trying to to send out multiple unsolicited responses (to a single or multiple clients?) based on some event happening.

The proxy logic makes use of separate responses - with the response from the upstream server being passed back at a later time to the requesting client.  The proxy logic does not use Async support to do this.

There is no reason I can think of for you to not send out unsolicited responses.  However the receiving client will need to know what to do with it (usually done by token matching as the request is not returned in the response).

Is what you are trying to do is to provide a composite data response for a resource request that has to get the information from different sources for each sub element of the composite data?
[...]

PR #1701, opened 2025-06-23
make async release after application callback conditional to enable multiple async triggers

mkicherer:
As discussed in #1696, this commit enables triggering an async object multiple times. I am not really happy with the commit message but I had no better idea so far.

In the example at the end of https://libcoap.net/doc/reference/develop/man_coap_async.html, it is noted that the async object is freed after the callback. Should I add a note there that this can be avoided by a `coap_async_set_delay` call?

Thank you!

### DIFF
--- a/man/coap_async.txt.in
+++ b/man/coap_async.txt.in
@@ -209,7 +209,10 @@ hnd_get_with_delay(coap_session_t *session,
                                (const uint8_t *)"done", NULL, NULL);
   coap_pdu_set_code(response, COAP_RESPONSE_CODE_CONTENT);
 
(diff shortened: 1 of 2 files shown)


mrdeep1:
I suggest you update the commit to be
````
Async: Support re-use of async with multiple triggers

Support re-use of async for multiple triggers by resetting the async delay.
````
Replace last comment line of man page async examples with
````
  /*
   * async is automatically removed by libcoap on return from this handler unless
   * coap_async_set_delay() is called.
   */
````

mkicherer:
I will handle this with an internal state machine that will trigger libcoap's async only once.

PR #1707, opened 2025-07-02
coap_net: Use platform-independent format specifier

kkrentz:
This PR corrects a format specifier in `coap_net.c`.


mrdeep1:
Thanks for fixing this.

PR #1721, opened 2025-07-29
Zephyr: improve POSIX API integration

sandro97git:
using zephyrs posix api should not require including headers from a zephyr specific location


mrdeep1:
@sandro97git Thanks for doing this work.

@fj-blanco Can you please check that this does not break your Zephyr work?

fj-blanco:
The [clients](https://github.com/fj-blanco/coap-zephyr-clients) work with this changes. Thank you @sandro97git for improving this.

mrdeep1:
@fj-blanco Thanks for confirming.

PR #1760, opened 2025-10-01
Fix HAVE_TIME_H and HAVE_SYS_TIME_H ifdefs

### LINKED ISSUE #1734: LWIP without OS-awareness produces a compile errors
This software library is a masterpiece, 😎 and we decided to use it under extreme conditions on a minimalist IoT controller. However, we encountered an unexpected problem: it's impossible to compile libcoap for the tiny STM32 with LWIP 2.1.2 but without OS-awareness (i.e., without FreeRTOS). 😔

The functions and constants `getaddrinfo()`, `AF_UNSPEC`, `AF_INET` etc. are undefined when a LWIP is used (`WITH_LWIP` = 1) but without an OS-awareness (`NO_SYS` = 1) and dependly without a Sockets API (`LWIP_SOCKET` = 0). The first compile error goes there: `coap_address.c:coap_resolve_address_info()`. Moreover the `struct addrinfo` can be also undefined by `LWIP_DNS` = 0.

As a quick and dirty workaround, the following code can be added to the user's  `coap_config.h` file:

```
#if WITH_LWIP && NO_SYS

#define AF_UNSPEC       0
#define AF_INET         2

#define getaddrinfo(nodname, servname, hints, res) \
   lwip_getaddrinfo(nodname, servname, hints, res)

typedef uint32_t in_addr_t;

struct in_addr {
  in_addr_t s_addr;
};

struct addrinfo {
  int               ai_flags;      /* Input flags. */
  int               ai_family;     /* Address family of socket. */
  int               ai_socktype;   /* Socket type. */
  int               ai_protocol;   /* Protocol of socket. */
  uint32_t          ai_addrlen;    /* Length of socket address. */
  struct sockaddr  *ai_addr;       /* Socket address of socket. */
  char             *ai_canonname;  /* Canonical name of service location. */
  struct addrinfo  *ai_next;       /* Pointer to next in list. */
};

struct sockaddr_in {
  uint8_t         sin_len;
  uint8_t         sin_family;
  uint16_t        sin_port;
  struct in_addr  sin_addr;
#define SIN_ZERO_LEN 8
  char            sin_zero[SIN_ZERO_LEN];
};

#endif /* WITH_LWIP && NO_SYS */
```
By the way, the `check_function_exists(getaddrinfo HAVE_GETADDRINFO)` call is used in the CMakeLists.txt file, but the result is not used anywhere.

I humbly hope that someone will be able to adapt the library for the above conditions in the right way.

raspopov:
If we correctly determine the presence of the `sys/time.h` and `time.h` files using the `HAVE_TIME_H` and `HAVE_SYS_TIME_H` macros, then the LWIP compilation will result in a duplicate function error. Perhaps previous versions of LWIP did not have these files, in which case everything would have worked correctly.

These changes add the necessary ifdefs and keep the optimal timer function for LWIP.

Partial fixes #1734.


raspopov:
```
pdu_parse_tcp_target.c:
include/coap3/coap_net.h:831:49: error: unknown type name ‘fd_set’
```
I don't quite understand why the error occurs especially considering my changes... 🤔

Missed `#include "coap_config.h"`?

mrdeep1:
> I don't quite understand why the error occurs especially considering my changes...

First issue is that `sys/select.h` is not included in AC_CHECK_HEADERS in `configure.ac` and so HAVE_SYS_SELECT_H was never set.

> Missed #include "coap_config.h"?

That is the second issue. However, applications should not need to be aware of `coap_config.h` which is used to build the libcoap library and may be different to what is available when doing multi platform / multi architecture builds.  It is not included in the built `include/coap3` directory (but files such as `coap_net.h` are) so a developer who is just using the library and include files can build their code..

`coap3/coap_defines.h` is available, but this only contains the COAP_ prefixed definitions.

As a general principal it is OK for platform based #if wrappers to be used in the `coap3/coap_*.h` files, but not based on whether a specific header file exists.

raspopov:
Thank you for your comment! However, could you please explain in more detail how this happened, as I don't see anything in my PR that would change the inclusion of the `<sys/select.h>` file and/or the `#ifdef HAVE_SYS_SELECT_H`?

mrdeep1:
> would change the inclusion of the <sys/select.h> file and/or the #ifdef HAVE_SYS_SELECT_H

If HAVE_SYS_SELECT_H was not defined, you then get the `fd_set `type issues.  Then the wrapping of `time.h` and `sys/time.h` have caused further issues.

In the way that the original file had a `_WIN32` wrapper, you would be better off adding a `WITH_LWIP` wrapper in `coap3/coap_net.h`.

raspopov:
What is the difference between these two tasks, “additional-tests (pull_request)” and “documentation (pull_request)”, and the others? Why do other tasks compile normally, but these do not?

mrdeep1:
Both of these CI tests are building applications that only use the public libcoap API, not the internal API.  This is dome by including `coap3/coap.h` and optionally `coap3/coap_defines.h` (the latter is not available for all platforms). ` coap.h` only includes the public #include files not the ones ending in `_internal.h` or `coap_libcoap_build.h`.

For the public API header files, it is problematic to use `#ifdef HAVE_` type wrappers for including additional header files which you have introduced to `coap_net.h`.  (The only other public header file coap_encode.h does use #ifndef HAVE_, but nowhere are these are defined.) 

As the applications (or their test runs) that are built in the 2 failing CI (linux based) tests include `coap.h` which includes `coap_net.h` and the `HAVE_`s are not defined you are getting these build errors as the wrapped include files are not getting included.

As mentioned before, if you are wanting to exclude some header files being included in a public API header file, the safe way to do this is to wrap them with the platform (or compiler) definitions, not by using the `HAVE_ `definitions which are generated only for the library build usage into` coap_config.h` (which is NOT a public API file).  In other words revert your changes back to using the `#ifndef _WIN32` approach method you replaced. 

To check your updated code is working as expected, on a linux based system before pushing the code changes

additional-tests
````
./autogen.sh
./configure --disable-tests --disable-documentation
make
make -C tests/oss-fuzz -f Makefile.ci check clean
````
documentation
````
./autogen.sh
./configure --disable-tests --enable-documentation
make -C man
man/examples-code-check man
````

raspopov:
I realized what was confusing me. The `sys/time.h` file includes the `sys/select.h` file, which is surprising, but true. So, even though the `sys/select.h` file in `coap_net.h` seems excluded by ifdef, it is still used because of the inclusion of `sys/time.h`. 

Additionally, when preparing the documentation, all example listings are extracted and compiled using a custom program. All of these errors are due to this nested compilation.

I hope I now understand the interesting mechanism of libcoap compilation and will soon release an update.

raspopov:
Clean PR: https://github.com/obgm/libcoap/pull/1768

PR #1761, opened 2025-10-01
Fix coap_resolve_address_info() for LWIP

### LINKED ISSUE #1734: LWIP without OS-awareness produces a compile errors

raspopov:
These changes add a basic implementation of the `coap_resolve_address_info()` function when compiling LWIP with `NO_SYS = 1`. This implementation is similar to those in CONTIKI and RIOT.

The function `gai_strerror()` was removed because it is not very important, yet it is missing from many implementations.

A CONTIKI compilation bug has also been fixed.

Fixes #1734.

### DIFF
--- a/src/coap_address.c
+++ b/src/coap_address.c
@@ -573,6 +573,8 @@ update_coap_addr_port(coap_uri_scheme_t scheme, coap_addr_info_t *info,
   }
 }
 
+#if defined(WITH_LWIP) && !(LWIP_DNS && LWIP_SOCKET && LWIP_COMPAT_SOCKETS)
+
 coap_addr_info_t *
 coap_resolve_address_info(const coap_str_const_t *address,
                           uint16_t port,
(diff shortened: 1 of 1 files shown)


src/coap_address.c
```diff
@@ -876,12 +951,24 @@ coap_resolve_address_info(const coap_str_const_t *address,
...
+                          uint16_t ws_port,
+                          uint16_t ws_secure_port,
+                          int ai_hints_flags,
+                          int scheme_hint_bits,
+                          coap_resolve_type_t type) {
   return NULL;
-#endif
 }
 
+#endif
```
mrdeep1, line 976: I think I prefer the #bad to be here so a new platform will throw up a compile warning rather that having to debug an address that is not parsing.

src/coap_address.c
```diff
@@ -643,7 +696,7 @@ coap_resolve_address_info(const coap_str_const_t *address,
   error = getaddrinfo(addrstr, NULL, &hints, &res);
 
   if (error != 0) {
-    coap_log_warn("getaddrinfo: %s: %s\n", addrstr, gai_strerror(error));
+    coap_log_warn("getaddrinfo: %s: %d\n", addrstr, error);
```
mrdeep1, line 700: My preference would be to still have `gai_strerror()` to translate the `getaddrinfo()` that will have more meaning to most developers.  It is also used elsewhere in the code.

I agree that some platforms do not support `gai_strerror()` and stub functions have had to be written.

mrdeep1 (commented):
Thanks for fixing this.

Please squash your merges into one.

PR #1762, opened 2025-10-01
Fix Doxygen under Windows

raspopov:
The CMakeLists.txt file calls Doxygen and redirects its output to `/dev/null` to suppress it. However, for Windows, you need to redirect it to `NUL:` otherwise, a build error will occur.


mrdeep1:

PR #1765, opened 2025-10-02
Fix Windows tests

raspopov:
In Windows, calling the `connect()` function for a socket requires a valid address. This causes the `t_session6()` test in the `coap_new_client_session()` function in `test_session.c` to fail.

The fix replaces `in6addr_any` with `in6addr_loopback`, as in the adjacent function call. This does not affect the functionality of the test.

As a result, the tests under Windows are completely successful.

### DIFF
--- a/tests/test_session.c
+++ b/tests/test_session.c
@@ -155,11 +155,10 @@ t_session6(void) {
 
   laddr.size = sizeof(struct sockaddr_in6);
   laddr.addr.sin6.sin6_family = AF_INET6;
-  laddr.addr.sin6.sin6_addr = in6addr_any;
+  laddr.addr.sin6.sin6_addr = in6addr_loopback;
   laddr.addr.sin6.sin6_port = htons(COAP_DEFAULT_PORT);
 
   coap_address_copy(&saddr, &laddr);
-  saddr.addr.sin6.sin6_addr = in6addr_loopback;
   saddr.addr.sin6.sin6_port = htons(20000);
 
   session = coap_new_client_session(ctx, &saddr, &laddr, COAP_PROTO_UDP);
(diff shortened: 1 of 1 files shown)


tests/test_session.c
```diff
@@ -155,11 +155,10 @@ t_session6(void) {
...
   laddr.addr.sin6.sin6_family = AF_INET6;
-  laddr.addr.sin6.sin6_addr = in6addr_any;
+  laddr.addr.sin6.sin6_addr = in6addr_loopback;
   laddr.addr.sin6.sin6_port = htons(COAP_DEFAULT_PORT);
 
   coap_address_copy(&saddr, &laddr);
-  saddr.addr.sin6.sin6_addr = in6addr_loopback;
   saddr.addr.sin6.sin6_port = htons(20000);
 
   session = coap_new_client_session(ctx, &saddr, &laddr, COAP_PROTO_UDP);
```
mrdeep1, line 164: Is not the underlying issue here that the `saddr `(server) and `laddr `(local) parameters should be changed in the `coap_new_client_session()` call and none of the other changes are required?

raspopov: 🤔 Then the port number assignment will probably have to be corrected. On the other hand, in my version, the resulting code is simpler and more compact. It also removes the repeated sequential assignment of the value to the variable `addr.addr.sin6.sin6_addr`. It seems that this incompatibility has already come up before, and someone has fixed it.

mrdeep1: I see that I had reversed this in #280 to support BSD builds.  Happy to merge as is.

raspopov: Thanks so much for your time. 😎

PR #1766, opened 2025-10-02
Add CUnit autodetection to CMake

raspopov:
Added a rules file for automatic detection of the [CUnit](https://cunit.sourceforge.net/) library, as well as support for running tests directly from CMake.

The CUnit tests run perfectly on both Windows and Debian.


mrdeep1:
Thanks for doing this. `cmake/FindCUnit.cmake` needs to be added to EXTRA_DIST in `Makefile.am` to make sure that the file is included in the distribution bundle.

mrdeep1:
Thanks for doing this.

PR #1768, opened 2025-10-03
Fix HAVE_TIME_H and HAVE_SYS_TIME_H ifdefs

raspopov:
If we correctly determine the presence of the `sys/time.h` and `time.h` files using the `HAVE_TIME_H` and `HAVE_SYS_TIME_H` macros, then the LWIP compilation will result in a duplicate function error. Perhaps previous versions of LWIP did not have these files, in which case everything would have worked correctly. This PR resolves all remaining incompatibilities with LWIP version 2.1.2, whether or not FreeRTOS is used.

These changes add the necessary ifdefs and keep the optimal timer function for LWIP.

Adds missed `#include <time.h>` lines to examples and documentation.

Adds the detection of the `sys/select.h` header to the CMake and autoconf.


include/coap3/coap_net.h
```diff
@@ -19,13 +19,10 @@
...
 #include <string.h>
-#ifndef _WIN32
-#ifdef HAVE_SYS_SELECT_H
+
+#if HAVE_SYS_SELECT_H || (!defined(WITH_LWIP) && !defined(WITH_CONTIKI) && !defined(RIOT_VERSION) && !defined(_WIN32))
 #include <sys/select.h>
-#endif /* HAVE_SYS_SELECT_H */
-#include <sys/time.h>
 #endif
-#include <time.h>
```
mrdeep1, line 28: In the spirit of not using HAVE_ conditional checks in the public API header files, I would prefer this to be
````
diff --git a/include/coap3/coap_net.h b/include/coap3/coap_net.h
index aa503eb5..e6d37ab1 100644
--- a/include/coap3/coap_net.h
+++ b/include/coap3/coap_net.h
@@ -19,13 +19,12 @@

 #include <stdlib.h>
 #include <string.h>
-#ifndef _WIN32
-#ifdef HAVE_SYS_SELECT_H
+
+#if !defined(WITH_LWIP) && !defined(WITH_CONTIKI) && !defined(RIOT_VERSION) && !defined(_WIN32)
 #include <sys/select.h>
-#endif /* HAVE_SYS_SELECT_H */
 #include <sys/time.h>
-#endif
 #include <time.h>
+#endif /* ! WITH_LWIP && ! WITH_CONTIKI && ! RIOT_VERSION && ! _WIN32 */

 #ifdef WITH_LWIP
 #include <lwip/ip_addr.h>
````
which then also puts back <time.h> and <sys/time.h> (suitably wrapped).

Then I believe a lot of the other changes are no longer needed.

As an FYI LwIP 2.1.2 is old compared to the general libcoap build which makes use of LwIP 2.2.1

raspopov: If I may, I would like to bring the following facts to your attention. The purpose of this PR is to eliminate "the toxic" headers `time.h` and `sys/time.h`, particularly `sys/time.h`, which is tied to POSIX sockets and includes `sys/select.h`. It is important to note that the libcoap library supports non-POSIX systems (for example LWIP without OS). In general, it is a bad idea to re-export system headers within library headers. The missing includes in the fixed examples demonstrate the disadvantages of the reverse approach. This is especially true considering that libcoap itself isolates "all time" within `coap_time.c` and does not use these includes itself.

raspopov: > As an FYI LwIP 2.1.2 is old compared to the general libcoap build which makes use of LwIP 2.2.1

I mentioned the old version of LwIP, 2.1.2, because that's the one I tested. The LwIP version is closely linked to the development tool; in this case, it is the Cube IDE. Even the latest version of the Cube IDE for our STM32 processors still contains this old version of LwIP.

mrdeep1: OK - lets eliminate the toxic `time.h` and` sys/time.h` issues.

However, I also want to get rid of the use of the HAVE_ conditional checks in the public API headers - in this case HAVE_SYS_SELECT.  So, an update to your code changes would be :-
````
diff --git a/include/coap3/coap_net.h b/include/coap3/coap_net.h
index ef19d055..310f4b84 100644
--- a/include/coap3/coap_net.h
+++ b/include/coap3/coap_net.h
@@ -20,9 +20,9 @@
 #include <stdlib.h>
 #include <string.h>

-#if HAVE_SYS_SELECT_H || (!defined(WITH_LWIP) && !defined(WITH_CONTIKI) && !defined(RIOT_VERSION) && !defined(_WIN32))
+#if !defined(WITH_LWIP) && !defined(WITH_CONTIKI) && !defined(RIOT_VERSION) && !defined(_WIN32)
 #include <sys/select.h>
-#endif
+#endif /* ! WITH_LWIP && ! WITH_CONTIKI && ! RIOT_VERSION && ! _WIN32 */

 #ifdef WITH_LWIP
 #include <lwip/ip_addr.h>
````

raspopov: Yes, I agree with you. I thought long and hard about whether to keep this definition for the same reason: the interface should not include implementation details, if possible. 🤔Unfortunately, this is hardly possible for a multi-platform library such as libcoap (you need to work hard on type erasure, and C++ is better suited for this).

However, I agree that this definition is not necessary for current configurations (it is more useful for hypothetical future implementations of LWIP/Contiki/Riot/Windows).

mrdeep1:
Thanks for doing this.

PR #1772, opened 2025-10-05
Fix compiler warnings

raspopov:
Fixed warnings of unused variable (for Release build) and possible use of initialized variable (for not so smart compilers).


mrdeep1:

PR #1775, opened 2025-10-06
Fix MSVC compilation (CMake-based)

raspopov:
Using the Ninja generator with the MSVC compiler leads to errors due to incorrect compiler call options. In CMake, the generator and the compiler are different entities.

This fix replaces the incorrect CMake variable with the correct one that describes the current MSVC-compatible compiler.


mrdeep1:
Thanks for fixing this.

PR #1776, opened 2025-10-06
Fix tinyDTLS compilation

raspopov:
The current code uses a dynamic array on the stack, a relatively rare compiler extension. For example, MSVC does not support it.

The proposed changes use a classic dynamic array on the heap, which greatly improves code compatibility. This will also enable libcoap to be compiled alongside tinyDTLS using the Microsoft compiler on Windows.

### DIFF
--- a/src/coap_tinydtls.c
+++ b/src/coap_tinydtls.c
@@ -673,11 +673,16 @@ verify_ecdsa_key(struct dtls_context_t *dtls_context COAP_UNUSED,
   int ret;
 
   if (t_context && t_context->setup_data.validate_cn_call_back) {
+    coap_address_t remote_addr;
+    get_session_addr(dtls_session, &remote_addr);
+    coap_session_t *c_session = coap_session_get_by_peer(t_context->coap_context,
+                                                         &remote_addr, dtls_session->ifindex);
+    if (!c_session)
+      return -3;
+
     /* Need to build asn.1 certificate - code taken from tinydtls */
     uint8 *p;
-    uint8 buf[DTLS_CE_LENGTH];
-    coap_session_t *c_session;
-    coap_address_t remote_addr;
+    uint8 *buf = coap_malloc_type(COAP_STRING, DTLS_CE_LENGTH);
 
     /* Certificate
      *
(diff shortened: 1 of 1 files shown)


src/coap_tinydtls.c
```diff
@@ -673,11 +673,16 @@ verify_ecdsa_key(struct dtls_context_t *dtls_context COAP_UNUSED,
...
+                                                         &remote_addr, dtls_session->ifindex);
+    if (!c_session)
+      return -3;
+
     /* Need to build asn.1 certificate - code taken from tinydtls */
     uint8 *p;
-    uint8 buf[DTLS_CE_LENGTH];
-    coap_session_t *c_session;
-    coap_address_t remote_addr;
+    uint8 *buf = coap_malloc_type(COAP_DTLS_CONTEXT, DTLS_CE_LENGTH);
```
mrdeep1: COAP_DTLS_CONTEXT may be of fixed size (in particular LwIP) and so may not be >= DTLS_CE_LENGTH.

I would use COAP_STRING here.

src/coap_tinydtls.c
```diff
@@ -693,16 +698,10 @@ verify_ecdsa_key(struct dtls_context_t *dtls_context COAP_UNUSED,
...
-
-    get_session_addr(dtls_session, &remote_addr);
-    c_session = coap_session_get_by_peer(t_context->coap_context,
-                                         &remote_addr, dtls_session->ifindex);
-    if (!c_session)
-      return -3;
     coap_lock_callback_ret(ret,
                            t_context->setup_data.validate_cn_call_back(COAP_DTLS_RPK_CERT_CN,
                                buf, p-buf, c_session, 0, 1, t_context->setup_data.cn_call_back_arg));
+    coap_free_type(COAP_DTLS_CONTEXT, buf);
```
mrdeep1: COAP_DTLS_CONTEXT may be of fixed size (in particular LwIP) and so may not be >= DTLS_CE_LENGTH.

I would use COAP_STRING here.

mrdeep1 (changes requested):
Thanks for raising this.

src/coap_tinydtls.c
raspopov: Thank you for the helpful explanation. I was wondering how memory pools differ...

mrdeep1 (approved):
Change looks good to me

PR #1777, opened 2025-10-06
Switch the CMake TinyDTLS build to the native way

raspopov:
These changes switch the TinyDTLS library build from Autotools to CMake, as support for CMake has now been added. The use of external libraries has been optimized. PkgConfig support has been added for library detection. The libraries' names were also fixed to use CMake-style naming.

🧐 I would like to note that there are problems compiling TinyDTLS under MSVC due to errors in `ext/tinydtls/dtl_debug.c` and `ext/tinydtls/CMakeLists.txt`. I will try to fix these errors upstream...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,21 +108,21 @@ set(DTLS_BACKEND
       STRING
       "\
 Name of the dtls backend, only relevant if `ENABLE_DTLS` is ON which is default. \
-Possible values: default, gnutls, openssl, wolfssl, tinydtls and mbedtls. \
+Possible values: default, gnutls, openssl, wolfssl, mbedtls and tinydtls. \
 If specified then this library will be searched and if found also used. \
 If not found then the cmake configuration will stop with an error. \
 If not specified, then cmake will try to use the first one found in the following order: \
-gnutls, openssl, wolfssl, tinydtls, mbedtls \
+gnutls, openssl, wolfssl, mbedtls, tinydtls \
     ")
 set_property(
   CACHE DTLS_BACKEND
   PROPERTY STRINGS
            default
+           gnutls
            openssl
            wolfssl
-           gnutls
-           tinydtls
-           mbedtls)
+           mbedtls
+           tinydtls)
 option(
   USE_VENDORED_TINYDTLS
   "compile with the tinydtls project in the submodule if on, otherwise try to find the compiled lib with find_package"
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -848,7 +802,6 @@ target_include_directories(
          $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include/>
          $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include/>
          $<INSTALL_INTERFACE:include/>
-         $<BUILD_INTERFACE:$<$<AND:$<BOOL:${COAP_WITH_LIBTINYDTLS}>,$<BOOL:${USE_VENDORED_TINYDTLS}>>:${CMAKE_BINARY_DIR}/include/tinydtls>>
```
mrdeep1, line 851: To be able to build with USE_VENDORED_TINYDTLS set to OFF. you need the following 2 lines here
````
+         $<$<BOOL:${COAP_WITH_LIBTINYDTLS}>:${TINYDTLS_INCLUDE_DIR}>
+         $<$<BOOL:${COAP_WITH_LIBTINYDTLS}>:${TINYDTLS_INCLUDE_DIR}/tinydtls>
````
It appears that the second line is needed as well when building the tinydtls' own subdirectories.

**Review by mrdeep1 — CHANGES_REQUESTED** [review]
Thanks for this.  A comment on a change plus this warning needs to be removed when USE_VENDORED_TINYDTLS is set to NO by changing all tinyDTLS to TinyDTLS in `cmake/FindTinyDTLS.cmake` .
````
CMake Warning (dev) at /usr/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (tinyDTLS)
  does not match the name of the calling package (TinyDTLS). 
````

File: CMakeLists.txt (hunk 1 above)
Comment (raspopov) [reply CMakeLists.txt:851]: Thank you for taking the time to review my code.
I already know how to do this in a more elegant way by making changes to `FindTinyDTLS.cmake`.

Comment (mrdeep1) [reply CMakeLists.txt:851]: > I already know how to do this in a more elegant way by making changes to FindTinyDTLS.cmake.

Great.  It would be good to get this properly fixed, so we can merge this PR.

Comment (raspopov) [reply CMakeLists.txt:851]: > To be able to build with USE_VENDORED_TINYDTLS set to OFF

We should probably add a CI job for this build option. As far as I understand, `tinydtl` does not yet have public packages for distributions. Is that correct?

And it seems reasonable to wait for this change to be adopted: https://github.com/eclipse-tinydtls/tinydtls/pull/263.

**Comment by raspopov** [comment]
Zephyr has been infused into libcoap so brutally... 🤔

File: cmake/Findwolfssl.cmake [hunk 2]
```diff
@@ -1,5 +1,5 @@
-# FindWolfSSL.cmake
-# -----------------
+# Findwolfssl
```
mrdeep1, line 1: Any reason as to why we are not staying with the other TLS library naming conventions - namely use wolfSSL and not the updated version of wolfssl?

wolfSSL refers to wolfSSL in all of their documentation.

CMakeLists.txt
mrdeep1: > We should probably add a CI job for this build option.

Makes sense.  TinyDTLS will need to git cloned, built and installed prior to libcoap in the CI.

> And it seems reasonable to wait for this change to be adopted: https://github.com/eclipse-tinydtls/tinydtls/pull/263.

That could take some time.  The CI job could be in a separate PR.

cmake/Findwolfssl.cmake
raspopov: Thanks for the good question. The WolfSSL library is built using CMake. It also supports CMake integration into other projects via configuration and targets. In these integration files, the library refers to itself as "wolfssl:wolfssl" (lowercase). For compatibility, I thought it would be better to use the library's self-name, even though libcoap does not use this type of integration yet.

mrdeep1:
Thanks for doing this tidy up work.

Issue #1778, opened 2025-10-07
Unexpected response delay after request caused server to use async when Q-Block method enabled


mkicherer:
Hello,

I noticed an unexpected delay for requests when the request causes the server to use separate/async response and COAP_BLOCK_TRY_Q_BLOCK is enabled. It looks to me like the client does not send an ACK for the received response to the Q-Block support test and that causes the response to the actual request to be delayed by the server. After a retransmit of the check response, the ACK is received by the server and it sends the actual response.

The client log:
```
Oct 07 15:29:08.170 DEBG ***[fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: new outgoing session
Oct 07 15:29:08.170 DEBG *  [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: netif: sent  247 bytes
Oct 07 15:29:08.170 DEBG timeout is set to 90 seconds
Oct 07 15:29:08.170 DEBG sending CoAP request:
Oct 07 15:29:08.170 DEBG Testing for Q-Block support
Oct 07 15:29:08.170 DEBG ** [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: mid=0x1190: delayed
Oct 07 15:29:08.172 DEBG *  [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: netif: recv   60 bytes
Oct 07 15:29:08.173 DEBG *  [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: netif: sent  279 bytes
Oct 07 15:29:08.176 DEBG *  [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: netif: recv  179 bytes
Oct 07 15:29:08.176 DEBG got psk_identity_hint: '...'
Oct 07 15:29:08.176 DEBG *  [fe80::17de:bafd:6df8:841d]:38335 <-> [fe80::20f:69ff:feff:d6c3]:5684 DTLS: netif: sent  132 bytes
[...]

PR #1780, opened 2025-10-09
add COAP_BLOCK_FORCE_Q_BLOCK

mkicherer:
As discussed in #1778, this adds a flag to force Q-Block method without checking for compatibility of the peer. This avoids an initial request of .well-known/core in environments where all peers are known to support the Q-Block method.

### DIFF
--- a/examples/coap-client.c
+++ b/examples/coap-client.c
@@ -593,7 +593,7 @@ usage(const char *program, const char *version) {
           "\t-K interval\tSend a ping after interval seconds of inactivity\n"
           "\t-L value\tSum of one or more COAP_BLOCK_* flag valuess for block\n"
           "\t       \t\thandling methods. Default is 1 (COAP_BLOCK_USE_LIBCOAP)\n"
-          "\t       \t\t(Sum of one or more of 1,2,4,8,16 and 32)\n"
+          "\t       \t\t(Sum of one or more of 1,2,4,8,16,32 and 512)\n"
           "\t-N     \t\tSend NON-confirmable message\n"
           "\t-O num,text\tAdd option num with contents text to request. If the\n"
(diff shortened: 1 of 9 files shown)


examples/coap-client.c
```diff
@@ -593,7 +593,7 @@ usage(const char *program, const char *version) {
           "\t-K interval\tSend a ping after interval seconds of inactivity\n"
           "\t-L value\tSum of one or more COAP_BLOCK_* flag valuess for block\n"
           "\t       \t\thandling methods. Default is 1 (COAP_BLOCK_USE_LIBCOAP)\n"
-          "\t       \t\t(Sum of one or more of 1,2,4,8,16 and 32)\n"
+          "\t       \t\t(Sum of one or more of 1,2,4,8,16,32 and 512)\n"
```
mrdeep1, line 596: `man/coap-client.txt.in` needs updating as well.

examples/coap-server.c
```diff
@@ -1585,7 +1585,7 @@ usage(const char *program, const char *version) {
           "\t       \t\tif the -A option is used\n"
           "\t-L value\tSum of one or more COAP_BLOCK_* flag valuess for block\n"
           "\t       \t\thandling methods. Default is 1 (COAP_BLOCK_USE_LIBCOAP)\n"
-          "\t       \t\t(Sum of one or more of 1,2,4 64, 128 and 256)\n"
+          "\t       \t\t(Sum of one or more of 1,2,4 64, 128, 256 and 512)\n"
```
mrdeep1, line 1588: `man/coap-server.txt.in` needs updating as well.

mrdeep1 (changes requested):
Thanks for doing this.  Looking good to me apart from a couple of comments.

PR #1781, opened 2025-10-10
Fix the MbedTSL compilation process under Windows

raspopov:
It adds the `strndup()` function, which is missing in Windows, as has been done in other parts of libcoap.


mrdeep1:
Thanks for fixing this.

PR #1782, opened 2025-10-10
Fix wolfSSL bug and MSVC compilation

raspopov:
When I was fixing MSVC compilation incompatibility due to a dynamic array on the stack in the `coap_crypto_aead_decrypt()` function, I accidentally noticed an incorrect calculation of the array size in the previous `coap_crypto_aead_encrypt()` function.


mrdeep1:
Thanks for finding, reporting and fixing this.

PR #1784, opened 2025-10-19
coap_encode.h: Check BSD is defined before evaluating it

ttefke:
Hello,

The BSD macro is evaluated even if it is not defined, leading to a build error. With this change, it is only evaluated if previously defined.

Best regards,
Tobias


mrdeep1:
Thanks for fixing this.

PR #1788, opened 2025-10-22
fixed memory leak in coap_send

mopsiok:
As the declaration of `coap_send` states, user must not manually release the PDU:
```c
* Sends a CoAP message to given peer. The memory that is
* allocated for the pdu will be released by coap_send().
* The caller must not use or delete the pdu after calling coap_send().
```

This PR fixes a memory leak for some edge cases, where `coap_send_internal` (called by `coap_send_lkd`, called by `coap_send`) returns without any additional memory release.


mrdeep1:
Thanks for troubleshooting and fixing this.

PR #1805, opened 2025-11-21
zephyr: mbedlts improvements

casaroli:
Add some improvements to make secure coap work on Zephyr 4.3.0 with mbedtls (not using Zephyr TLS protocol options).

```sh
west build -p always -b pico_plus2/rp2350b/m33/w libcoap/examples/zephyr/client-src --extra-conf libcoap-mbedtls.conf
```


mrdeep1:
Thanks for raising and working on this.

Fixing the pre-commit required changes should fix the the CI pre-commit, documentation and distribution builds.

In terms of the zephyr build, I would prefer that the generic MbedTLS (and WolfSSL) library setup configuration files were kept within the libcoap/zephyr directory for anyone to easily use rather then being in the libcoap/examples/zephyr directory.

casaroli:
Thank you for you comments.

> I would prefer that the generic MbedTLS (and WolfSSL) library setup configuration files were kept within the libcoap/zephyr directory for anyone to easily use rather then being in the libcoap/examples/zephyr directory

You mean the `libcoap_mbedtls.conf`? I moved it to the example directory because I believe this is "application configuration". I.e. if the application developer includes the `libcoap_mbedtls.conf`, then the application will have different build time configuration (include coaps support, mbedtls, etc).

In zephyr, the extra configuration file it is usually used (in other examples) as

```sh
build -p always -b <board> libcoap/examples/zephyr/client-src --extra-conf libcoap_mbedtls.conf
```

the config file is relative to the application directory. So I thought it was good idea to move it there, to simplify the west command on a standard Zephyr build.

What do you think?

I guess we could still move the file to `zephyr` and pass `--extra-conf ../../../zephyr/libcoap_mbedtls` to west if you like it better.

Or we could just enable mbedtls by default in Zephyr (i.e. move the stuff in `libcoap_mbedtls.conf` to `prj.conf` and push for a more secure world 😅 

Or just keep it like I suggested so we can simplify the `west` command and harmonize with how usually application extra configs are handled in examples.

Whatever you prefer, I can update the documentation to match.

casaroli:
I have applied the pre-commit and fixed the Makefile.

casaroli:
Ok this one is tricky. Zephyr mbedtls does not expose `mbedtls_version_get_number`. We can work around this by setting some dummy number here or wait for Zephyr to merge a PR to make it expose that function.

What do you think?

mrdeep1:
> Zephyr mbedtls does not expose mbedtls_version_get_number()

Actually it does if MBEDTLS_VERSION_C is defined. This was one of the reasons for having `zephyr/config-mbedtls-libcoap.h` in the `zephyr/` directory which you have deleted as it defines all the needed MBedTLS definitions to allow libcoap to work based on how MBedTLS is configured (via `Kconfig `and `prj.conf`).

That said, I am not a Zephyr expert and there may be better ways to do things.

casaroli:
Lets try this first: https://github.com/zephyrproject-rtos/zephyr/pull/99923

Please, bear with me on this 😅 

Thank you for your time.

mrdeep1:
> Lets try this first: https://github.com/zephyrproject-rtos/zephyr/pull/99923

The challenge  here is that we can currently build for MBedTLS, but builds will fail if someone is using a release version of Zephyr that does not include this fix.

casaroli:
can we just set this value as `unknown` if zephyr version is too low?

what is the earliest zephyr version you want to support?

mrdeep1:
> what is the earliest zephyr version you want to support?

Certainly base 4.3.0 version, and ideally a couple of versions earlier.

casaroli:
@mrdeep1 hello, can you please approve the workflow to be run so we can see if this works?

I figured out we can pass the mbedtls config parameters with `CMAKE_C_FLAGS` so we don't need to specify a manual mbedtls configuration.

I also notice you added `MBEDTLS_SSL_COOKIE_C` in the file I deleted, Zephyr should include it automatic by default if `CONFIG_MBEDTLS_SSL_PROTO_DTLS` is set, however we might need to set it like we did with `MBEDTLS_VERSION_C` for older versions of Zephyr.

mrdeep1:
Separately, it looks like there has been a zephyr update that is currently breaking existing libcoap builds which I am looking at.

casaroli:
> Separately, it looks like there has been a zephyr update that is currently breaking existing libcoap builds which I am looking at.

I suggest we try to use standard zephyr mbedtls configuration instead of providing our own mbedtls-config file.

This way, we can reduce the risk of breaking in the future (Zephyr already knows how to build mbedtls with DTLS support, we just need to set the proper zephyr configuration and it will configure mbedtls properly).

If we need extra mbedtls options not exported by Zephyr (like `MBEDTLS_VERSION_C`), we can just define them in the `CMAKE_C_FLAGS` and it will force-enable it. I don't think we need to explicitly enable the other options added to our custom mbedtls config file.

mrdeep1:
> I suggest we try to use standard zephyr mbedtls configuration instead of providing our own mbedtls-config file.

Keeping things simple is always one of my goals.

casaroli:
Since the build is targetting current zephyr dev (4.3.99) they deprecated this:

```
warning: Deprecated symbol XOPEN_STREAMS is enabled.
```

I will update to using `CONFIG_XSI_STREAMS` instead.

mrdeep1:
On something like an ubuntu host, I suggest you
````
cd examples/zephyr
make clean
make
````
to debug the vagaries of the zephyr build before doing another push.  If `zephyrproject/` in not installed in `examples/zephyr` directory, the latest version will get pulled down.

**Comment by casaroli** [comment]
> On something like an ubuntu host, I suggest you
> 
> ```
> cd examples/zephyr
> make clean
> make
> ```
> 
> to debug the vagaries of the zephyr build before doing another push. If `zephyrproject/` in not installed in `examples/zephyr` directory, the latest version will get pulled down.

Oh thank you!. I had a different version of zephyr in my zephyrproject, I was wondering why I was getting different results.

On another topic, do you want to keep trying to build against zephyr-next (i.e. 4.3.99 -- main branch)? It looks like it breaks constantly. Maybe it is better to target latest zephyr official releases (currently 4.3.0) instead of the development branch?

We can keep a draft PR open so we can keep working on the fixes ahead of the next release, so when they release (in march) we will be faster to migrate. I would not try to enforce that libcoap compiles cleanly with the active zephyr development branch on each PR.

WDYT?

**Comment by mrdeep1** [comment]
Certainly things are getting messier when building against the latest zephyr code.  I have just found that the libcoap develop branch does not enable PSK support - and if enabling it, then there is a mismatch in supported algorithms (not yet tracked down which).  This does need to get fixed.

At some point, the latest zephyr code will need to get supported, but I am going in the direction we support (by default) building against 4.3.0.

**Comment by mrdeep1** [comment]
I suggest you try apply this PR's simplification against #1877, which then should make things simpler when coming to support new Zephyr stable releases.

**Comment by casaroli** [comment]
> and if enabling it, then there is a mismatch in supported algorithms (not yet tracked down which). This does need to get fixed.

I had it fixed at some point in this PR (with Zephyr 4.3.0).

**Comment by mrdeep1** [comment]
> I had it fixed at some point in this PR (with Zephyr 4.3.0).

You can go back over the hashes of the various pushes (as recorded in this PR) to see if you can find your fixed version and work from there.

**Comment by casaroli** [comment]
@mrdeep1 I have dropped the last commit and rebased this against develop, and the other-build worked locally for me.

I will still test PSK but I suspect it is working.

**Comment by mrdeep1** [comment]
Building locally, the ./coap-client works (using DTLS), but there is a complier issue with this reported
````
/home/jon/libcoap/src/coap_mbedtls.c: In function ‘coap_get_tls_library_version’:
/home/jon/libcoap/src/coap_mbedtls.c:3448:21: warning: implicit declaration of function ‘mbedtls_version_get_number’; did you mean ‘mbedtls_timing_get_timer’? [-Wimplicit-function-declaration]
 3448 |   version.version = mbedtls_version_get_number();
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                     mbedtls_timing_get_timer
[1387/1389] Linking C executable zephyr/zephyr.elf
````

**Comment by mrdeep1** [comment]
I think you need to do this
````
diff --git a/examples/zephyr/Makefile b/examples/zephyr/Makefile
index 4344ab74..2016fd30 100644
--- a/examples/zephyr/Makefile
+++ b/examples/zephyr/Makefile
@@ -60,7 +60,8 @@ coap-client: client-src/src/main.c client-src/* ../../src/*.c ../../include/*/*
                $${LIBCOAP_DIR}/examples/zephyr/client-src -- \
                -DCONF_FILE=prj.conf \
                -DEXTRA_ZEPHYR_MODULES=$${LIBCOAP_DIR} \
-               -DEXTRA_CONF_FILE=libcoap-mbedtls.conf)
+               -DEXTRA_CONF_FILE=libcoap-mbedtls.conf \
+               -DCMAKE_C_FLAGS=-DMBEDTLS_VERSION_C)
        @cp -f $(ZEPHYR_PROJECT)/zephyr/build/zephyr/zephyr.exe coap-client
        @rm -rf $(ZEPHYR_PROJECT)/zephyr/build

@@ -73,8 +74,8 @@ coap-cpp-test: test-cpp-src/src/test.cpp test-cpp-src/* ../../src/*.c ../../incl
        west build -p always -b native_sim \
                $${LIBCOAP_DIR}/examples/zephyr/test-cpp-src -- \
                -DCONF_FILE=prj.conf \
-               -DEXTRA_ZEPHYR_MODULES=$${LIBCOAP_DIR})
-               -DCMAKE_C_FLAGS=-DMBEDTLS_VERSION_C
+               -DEXTRA_ZEPHYR_MODULES=$${LIBCOAP_DIR} \
+               -DCMAKE_C_FLAGS=-DMBEDTLS_VERSION_C)
        @cp -f $(ZEPHYR_PROJECT)/zephyr/build/zephyr/zephyr.exe coap-cpp-test
        @rm -rf $(ZEPHYR_PROJECT)/zephyr/build
 clean:
````

**Comment by casaroli** [comment]
Looks like this is needed only for coap-client not cpp-test. I am building locally now

**Comment by casaroli** [comment]
Oh, the change from `native_sim` to `native_sim/native/64` slipped in. If we want this we need it to be in separate commit, so I will just remove it now.

**Comment by casaroli** [comment]
@mrdeep1 I am sorry for all the wasted CI time.

Hey at least we are not using AI 🤣

File: zephyr/CMakeLists.txt [hunk 1]
```diff
@@ -113,6 +113,50 @@ if(CONFIG_LIBCOAP)
...
+set(HAVE_SELECT 1)
+set(HAVE_SOCKET 1)
+set(HAVE_STRCASECMP 1)
+set(HAVE_PTHREAD_MUTEX_LOCK 1)
+set(HAVE_GETADDRINFO 1)
+set(HAVE_STRNLEN 1)
+set(HAVE_STRRCHR 1)
+set(HAVE_GETRANDOM 1)
+set(HAVE_RANDOM 1)
+set(HAVE_IF_NAMETOINDEX 1)
```
mrdeep1: I think I would prefer these to be indented.

mrdeep1 (commented):
At some point these 3 commits need to be merged.

zephyr/CMakeLists.txt
casaroli: done.

casaroli:
Most of these defines are not used by libcoap, and it looks like `check_include_file` is problematic for Zephyr and/or cross compiling.

I would like to try (in a separate PR) to get rid of these unnecessary build-time configuration and the `check_include_file`.

mrdeep1:
For checking compilation warnings in the local build environment, it would be halpful to add in -Werror os per
````
diff --git a/examples/zephyr/Makefile b/examples/zephyr/Makefile
index fc0c7470..c61ebf76 100644
--- a/examples/zephyr/Makefile
+++ b/examples/zephyr/Makefile
@@ -60,7 +60,7 @@ coap-client: client-src/src/main.c client-src/* ../../src/*.c ../../include/*/*
                $${LIBCOAP_DIR}/examples/zephyr/client-src -- \
                -DCONF_FILE=prj.conf \
                -DEXTRA_ZEPHYR_MODULES=$${LIBCOAP_DIR} \
-               -DCMAKE_C_FLAGS=-DMBEDTLS_VERSION_C \
+               -DCMAKE_C_FLAGS="-DMBEDTLS_VERSION_C -Werror" \
                -DEXTRA_CONF_FILE=libcoap-mbedtls.conf)
        @cp -f $(ZEPHYR_PROJECT)/zephyr/build/zephyr/zephyr.exe coap-client
        @rm -rf $(ZEPHYR_PROJECT)/zephyr/build
@@ -74,6 +74,7 @@ coap-cpp-test: test-cpp-src/src/test.cpp test-cpp-src/* ../../src/*.c ../../incl
        west build -p always -b native_sim \
                $${LIBCOAP_DIR}/examples/zephyr/test-cpp-src -- \
                -DCONF_FILE=prj.conf \
+               -DCMAKE_C_FLAGS="-Werror" \
                -DEXTRA_ZEPHYR_MODULES=$${LIBCOAP_DIR})
        @cp -f $(ZEPHYR_PROJECT)/zephyr/build/zephyr/zephyr.exe coap-cpp-test
        @rm -rf $(ZEPHYR_PROJECT)/zephyr/build
````

casaroli:
done

mrdeep1:
> I would like to try (in a separate PR) to get rid of these unnecessary build-time configuration and the check_include_file.

Yes, it looks like a review is needed for the actual included header files in both `configure.ac` and `CMakeLists.txt` and unused header files (not anywhere in the lobcoap code) removed - but used header files should be suitably wrapped.

I still think we need `check_include_file `for the general build case, not necessarily for the cross compile case if it is not working properly - or to teach the cross compiler to NOT look in the host's include directories when testing `check_include_file`.

casaroli:
I will push a PR after this to propose the removal of  some of the most obvious unnecessary checks

mrdeep1:
For this PR, if you can merge the 4 commits into 1 as there are all related, I will then get this PR merged into develop.

Thanks for all of your help on  this one.

casaroli:
done. thank you for your time.

PR #1813, opened 2025-12-01
Fix CA cert loading failure with buffer-based certs when filesystem is disabled

danielriege:
## Problem

  Buffer-based CA certificates (`COAP_PKI_KEY_DEF_PEM_BUF`, `COAP_PKI_KEY_DEF_DER_BUF`)
  fail to load when `MBEDTLS_FS_IO` is not defined, even after successful parsing.

  The code unconditionally enters the root CA block (line 892) which returns
  `COAP_DEFINE_FAIL_NOT_SUPPORTED` when filesystem support is unavailable,
  failing the entire setup.

  ## Solution

  Added `done_ca_cert_buf` flag to skip the root CA block when a CA certificate
  has already been loaded from a buffer, preventing spurious failures in
  non-filesystem environments.

### DIFF
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -600,6 +600,7 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
   int ret;
   int done_private_key = 0;
   int done_public_cert = 0;
+  int done_ca_cert_buf = 0;
   uint8_t *buffer;
   size_t length;
 
@@ -860,6 +861,7 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
                                       &key, role, ret);
       }
       mbedtls_ssl_conf_ca_chain(&m_env->conf, cacert, NULL);
+      done_ca_cert_buf = 1;
       break;
     case COAP_PKI_KEY_DEF_RPK_BUF: /* define ca */
       return coap_dtls_define_issue(COAP_DEFINE_KEY_CA,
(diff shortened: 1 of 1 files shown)


mrdeep1:
The pre-commit (and hence documentation) check is complaining about the incorrect indent levels for the appropriate block of code which needs to be fixed.

However, would this not be a better fix so that if Root CAs or Trust Store are defined, only then complain there is an issue?
````
diff --git a/src/coap_mbedtls.c b/src/coap_mbedtls.c
index a8ef5467..064f7b85 100644
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -947,10 +947,11 @@ setup_pki_credentials(mbedtls_x509_crt *cacert,
     }
   }
 #else /* ! MBEDTLS_FS_IO */
-  (void)m_context;
-  return coap_dtls_define_issue(COAP_DEFINE_KEY_ROOT_CA,
-                                COAP_DEFINE_FAIL_NOT_SUPPORTED,
-                                &key, role, -1);
+  if (m_context->root_ca_file || m_context->root_ca_path || m_context->trust_store_defined) {
+    return coap_dtls_define_issue(COAP_DEFINE_KEY_ROOT_CA,
+                                  COAP_DEFINE_FAIL_NOT_SUPPORTED,
+                                  &key, role, -1);
+  }
 #endif /* ! MBEDTLS_FS_IO */

 #if defined(MBEDTLS_SSL_SRV_C)
````

mrdeep1:
Does #1832 work for you?

Issue #1810, opened 2025-11-28
The asymmetry between block transfer sending and receiving APIs makes it difficult for the libcoap library to function effectively on memory-constrained devices


Kev-Jiang:
**Is your feature request related to a problem? Please describe.**

In memory-constrained environments using libcoap's automatic block-wise transfers, I've encountered a significant limitation. While the receiving side can process blocks individually through callbacks (coap_register_block_data_handler), the sending side requires loading the entire data into memory at once with coap_add_data_large_* functions. This creates memory problems when working with large files that exceed available RAM.

**Describe the solution you would like**

I would like libcoap to provide a callback mechanism for the sending side similar to what exists for the receiving side. This would allow the implementation of a function that gets called to provide each block of data as needed during transmission, rather than requiring the entire payload to be in memory. Ideally, libcoap would still handle the block-wise protocol details automatically while using this callback to fetch data blocks on demand.

**Describe alternatives you have considered**

1. Manual block-wise transfers - implementing the block option logic myself, but this increases complexity and introduces potential for errors in the block-wise protocol implementation.

2. Chunking files manually and sending multiple separate CoAP messages, but this loses the benefits of block-wise transfers.

**Additional context**

This feature would be particularly valuable for IoT and embedded systems that need to transfer files or large payloads while maintaining minimal memory footprint. The asymmetry between receiving and sending capabilities in the current API creates unnecessary constraints for memory-limited devices.

anyc:
Hello,

I am currently investigating the same issue. I think I have a rough idea what we could do. As far as I understand the code, the functions `coap_add_data_large_internal` and `coap_handle_request_send_block` are of interest here. The former is responsible for adding the first block and the latter for the remaining blocks. In both cases, `coap_add_data` is called in the end. Maybe we could add a callback function to `coap_pdu_t` that will be called in this function?

However, I cannot say I understand all the code there. It is a bit difficult to follow for me.

mrdeep1:
I am thinking of a new function `coap_add_data_large_request_app()` (or similar) where a callback is supplied to get the data (data, length, offset) instead of providing the data for the` coap_add_data_large_request()` call.  The callback needs an additional parameter which could just be the `session `so that the app needs to know how / where to get the data.

anyc:
[...]

PR #1814, opened 2025-12-01
add coap_context_register_payload_transmit_callback

mkicherer:
This is a proof-of-concept patch for the introduction of an application callback that provides the data for the next block during a transmission in order to avoid keeping the complete payload in the RAM during the whole transfer. See also #1810.


mkicherer:
I only did the bare minimum to get it working so far. Checks will likely fail. I am also not sure if it is a good idea to calculate the data index using `data - pdu->lg_xmit->data_info->data` in `coap_add_data`.

mrdeep1:
Thanks for starting on this work.  #1815 is an alternative solution.

PR #1840, opened 2025-12-18
release lock during sleep on WIN32 and call second select only on demand

mkicherer:
Hello,

we noticed delays that only happen on Windows platforms. I remembered that only on Windows, the select call is omitted if no FDs are selected and then saw that we sleep instead without releasing the lock temporarily. So no other libcoap functions can execute during the selected timeout period. This patch releases the lock during the `sleep` and also avoids the second select call if no FDs are selected to avoid spurious "invalid argument" errors again.

I also replaced the `goto all_over` after the sleep, as I think the remaining code in the function might still be needed but I am not sure about this. I could not explain to myself why the second `select` call is needed in the `COAP_THREAD_SAFE` case. At least it should not hurt to execute the remainder.

Thank you!

### DIFF
--- a/src/coap_io_posix.c
+++ b/src/coap_io_posix.c
@@ -327,17 +327,19 @@ coap_io_process_with_fds_lkd(coap_context_t *ctx, uint32_t timeout_ms,
         milliseconds = 2000;
       }
       if (milliseconds) {
+        coap_lock_unlock();
         Sleep(milliseconds);
+        coap_lock_lock(return -1);
       }
     }
-    goto all_over;
+    result = 0;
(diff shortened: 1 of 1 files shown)


mrdeep1:
In doing the unlock / lock wrapper to` Sleep()` , then another thread can, say, make a new client connection, so the `nfds `and the `fds `state can change (and even do a` coap_io_process()`).  So, if there is multi-threaded support, once the `Sleep()` has finished and the global lock relocked, the current state of fds etc. needs to get rechecked.  So you have done the right thing here.

PR #1842, opened 2025-12-18
coap_io_process_with_fds_lkd: use local fd_set to avoid conflicts whe…

mkicherer:
Edit: Sorry, accidentally pressed enter.

When executing libcoap under WIN32 (in this particular case using wine), I noticed rare (and cryptic) errors that seem to be caused by select/poll. After adding some debug output, I saw that this function is called in parallel for the same context, e.g., if one thread is sleeping in the `select()` call. As both are using `ctx->readfds` and others, asserts seem to fail as one thread could modify the FD list while the other is using it. With this patch, the `fd_set` variables are thread-local and I did not see this error anymore so far.

I am not sure if this is the best way to solve it. What do you think?

### DIFF
--- a/src/coap_io_posix.c
+++ b/src/coap_io_posix.c
@@ -164,7 +164,9 @@ coap_io_process_with_fds(coap_context_t *ctx, uint32_t timeout_ms,
 static unsigned int
 coap_io_prepare_fds(coap_context_t *ctx,
                     int enfds, fd_set *ereadfds, fd_set *ewritefds,
-                    fd_set *eexceptfds) {
+                    fd_set *eexceptfds, fd_set *readfds, fd_set *writefds,
+                    fd_set *exceptfds
+                   ) {
   coap_session_t *s, *stmp;
   unsigned int max_sockets = sizeof(ctx->sockets) / sizeof(ctx->sockets[0]);
   coap_fd_t nfds = 0;
@@ -198,36 +200,36 @@ coap_io_prepare_fds(coap_context_t *ctx,
   }
 #endif /* COAP_CLIENT_SUPPORT */
   if (ereadfds) {
-    ctx->readfds = *ereadfds;
+    *readfds = *ereadfds;
     nfds = enfds;
   } else {
-    FD_ZERO(&ctx->readfds);
+    FD_ZERO(readfds);
   }
   if (ewritefds) {
-    ctx->writefds = *ewritefds;
+    *writefds = *ewritefds;
     nfds = enfds;
   } else {
-    FD_ZERO(&ctx->writefds);
+    FD_ZERO(writefds);
   }
   if (eexceptfds) {
-    ctx->exceptfds = *eexceptfds;
+    *exceptfds = *eexceptfds;
     nfds = enfds;
   } else {
-    FD_ZERO(&ctx->exceptfds);
+    FD_ZERO(exceptfds);
   }
   for (i = 0; i < ctx->num_sockets; i++) {
     if (ctx->sockets[i]->fd + 1 > nfds)
       nfds = ctx->sockets[i]->fd + 1;
     if (ctx->sockets[i]->flags & COAP_SOCKET_WANT_READ)
-      FD_SET(ctx->sockets[i]->fd, &ctx->readfds);
+      FD_SET(ctx->sockets[i]->fd, readfds);
     if (ctx->sockets[i]->flags & COAP_SOCKET_WANT_WRITE)
-      FD_SET(ctx->sockets[i]->fd, &ctx->writefds);
+      FD_SET(ctx->sockets[i]->fd, writefds);
 #if !COAP_DISABLE_TCP
     if (ctx->sockets[i]->flags & COAP_SOCKET_WANT_ACCEPT)
-      FD_SET(ctx->sockets[i]->fd, &ctx->readfds);
+      FD_SET(ctx->sockets[i]->fd, readfds);
     if (ctx->sockets[i]->flags & COAP_SOCKET_WANT_CONNECT) {
-      FD_SET(ctx->sockets[i]->fd, &ctx->writefds);
-      FD_SET(ctx->sockets[i]->fd, &ctx->exceptfds);
+      FD_SET(ctx->sockets[i]->fd, writefds);
+      FD_SET(ctx->sockets[i]->fd, exceptfds);
     }
 #endif /* !COAP_DISABLE_TCP */
   }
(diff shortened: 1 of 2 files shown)


mrdeep1:
The principal of doing this seems to be fine.

Pre-commit formatting errors need to be corrected.
I don't think that `coap_context_t` still needs to have `readfds`, `writefds `and `exceptfds`.
Other compilation errors due to compile time options need to be fixed.

My only concern is that for limited stack systems `3 * fd_set` as local stack variables may be problematic.  See `COAP_CONSTRAINED_STACK`.

mrdeep1:
I think that the `ctx->readfds` etc are suitably protected - hence the second set up of `ctx->readfds` etc. before calling `select()` for the second time (if thread safe enabled) so that the correct bits are set when deciding what to read/write.

However, for server endpoints, there may be an issue, fixed by (mirroring what is done in `coap_io_prepare_io_lkd()`)
````
diff --git a/src/coap_io_posix.c b/src/coap_io_posix.c
index 307af941..0555bcb4 100644
--- a/src/coap_io_posix.c
+++ b/src/coap_io_posix.c
@@ -181,7 +181,7 @@ coap_io_prepare_fds(coap_context_t *ctx,
     }
     SESSIONS_ITER(ep->sessions, s, stmp) {
       if (s->sock.flags & (COAP_SOCKET_WANT_READ|COAP_SOCKET_WANT_WRITE)) {
-        if (ctx->num_sockets < max_sockets)
+        if (ctx->num_sockets < max_sockets && !(s->sock.flags & COAP_SOCKET_SLAVE))
           ctx->sockets[ctx->num_sockets++] = &s->sock;
       }
     }
@@ -192,7 +192,7 @@ coap_io_prepare_fds(coap_context_t *ctx,
     if (s->sock.flags & (COAP_SOCKET_WANT_READ |
                          COAP_SOCKET_WANT_WRITE |
                          COAP_SOCKET_WANT_CONNECT)) {
-      if (ctx->num_sockets < max_sockets)
+      if (ctx->num_sockets < max_sockets && !(s->sock.flags & COAP_SOCKET_SLAVE))
         ctx->sockets[ctx->num_sockets++] = &s->sock;
     }
   }
````

**Comment by mkicherer** [comment]
I fixed the build error and added your patch.

I guess I could a macro that either uses ctx->readfds or the local readfds depending on `COAP_CONSTRAINED_STACK`. Shall I add this?

**Comment by mrdeep1** [comment]
> I guess I could a macro that either uses ctx->readfds or the local readfds depending on 
> COAP_CONSTRAINED_STACK. Shall I add this?

I still think that `ctx->readfds` etc. are correctly (re-)set up for actual usage when two threads are in `coap_io_process_lkd()` with one of them waiting on a `Sleep()` or first `select()` and then continuing when the other thread has made some changes to the state of the fds.

But yes, I guess a Macro could be used.

**Comment by mrdeep1** [comment]
I suspect you need to rebase the PR against the latest develop to fix the RIOT build issue.

**Comment by mkicherer** [comment]
My assumption is that the `select` implementation does not like that the FD_SET that was set when entering the syscall is different when it leaves. At least the error did not occur anymore so far. But maybe I just have to wait longer.

**Comment by mrdeep1** [comment]
> My assumption is that the select implementation does not like that the FD_SET that was set...

Good point,  potential for issues here.

File: src/coap_io_posix.c [hunk 1]
```diff
@@ -176,7 +178,7 @@ coap_io_prepare_fds(coap_context_t *ctx,
 
   LL_FOREACH(ctx->endpoint, ep) {
     if (ep->sock.flags & (COAP_SOCKET_WANT_READ | COAP_SOCKET_WANT_WRITE | COAP_SOCKET_WANT_ACCEPT)) {
-      if (ctx->num_sockets < max_sockets)
+      if (ctx->num_sockets < max_sockets && !(s->sock.flags & COAP_SOCKET_SLAVE))
```
mrdeep1: The endpoint socket will never be a slave, but sessions hanging off the endpoint will (for UDP/DTLS) will use the same socket as the endpoint, so do not need an additional sock included for each individual session.

So, this change here is not needed.

mrdeep1 (changes requested):
In general looking good.  However, because of the constrained stacks, we need to somehow handle not having `readfds`, `writefds `and `exceptfds ` in the stack for constrained stacks.

Unable to use static variables protected by `global_lock `as `global lock` is unlocked for `select()` and hence allowing multiple use.  Use of a separate mutex protection wont work as you are likely to get deadlocks.

mkicherer:
Do these platforms maybe have thread local storage?

src/coap_io_posix.c
```diff
@@ -247,7 +249,12 @@ coap_io_process_with_fds_lkd(coap_context_t *ctx, uint32_t timeout_ms,
   int result;
   unsigned int i;
 #endif /* ! COAP_EPOLL_SUPPORT */
-
+#if ! COAP_CONSTRAINED_STACK && ! COAP_EPOLL_SUPPORT
+  fd_set readfds, writefds, exceptfds;
+#define FD(fd) (fd)
+#else
+#define FD(fd) (ctx->(fd))
```
mrdeep1: To prevent future naming clashes, I think it best to have these as `COAP_FD()` rather than `FD()`.

I guess overall, I was thinking more of COAP_FD_READFDS, COAP_FD_WRITEFDS and COAP_FD_ACCEPTFDS, or even  COAP_READFDS, COAP_WRITEFDS and COAP_ACCEPTFDS for readability,

mrdeep1:
> Do these platforms maybe have thread local storage?

Possibly.  In the cases I know, each thread is allocated a maximum stack size based on the fact there is limited RAM and it is easy to hit this limit.

mkicherer:
As far as I know the "PC" operating systems use a separate memory region for the TLS, not the stack. But I have almost no experience with a microcontroller/realtime OS.

src/coap_io_posix.c
```diff
@@ -247,7 +249,16 @@ coap_io_process_with_fds_lkd(coap_context_t *ctx, uint32_t timeout_ms,
...
+#if ! COAP_CONSTRAINED_STACK && ! COAP_EPOLL_SUPPORT
+  fd_set readfds, writefds, exceptfds;
+#define COAP_READFDS readfds
+#define COAP_WRITEFDS writefds
+#define COAP_EXCEPTFDS exceptfds
+#else
+#define COAP_READFDS (ctx->readfds)
+#define COAP_WRITEFDS (ctx->writefds)
+#define COAP_EXCEPTFDS (ctx->exceptfds)
+#endif /* ! COAP_CONSTRAINED_STACK */
```
mrdeep1, line 260: This should include && ! COAP_EPOLL_SUPPORT.

If this #if block was to be moved up 2 lines to be inside the previous `#endif /* ! COAP_EPOLL_SUPPORT */`, then things would be cleaner.

mrdeep1:
Approved with the changes so far as there is no change to functionality or memory usage for constrained stack systems.

PR #1856, opened 2025-12-29
coap_uri.c: Add NULL checks after coap_new_optlist() calls

Oblivionsage:
`coap_new_optlist()` can return NULL on allocation failure. In `coap_path_into_optlist()` and `coap_query_into_optlist()`, the return value is passed directly to `coap_replace_percents()` without NULL check, leading to a potential NULL dereference on allocation failure.

This adds NULL checks in 4 places:
- `coap_path_into_optlist()`: 2 occurrences
- `coap_query_into_optlist()`: 2 occurrences

### DIFF
--- a/src/coap_uri.c
+++ b/src/coap_uri.c
@@ -782,6 +782,9 @@ coap_path_into_optlist(const uint8_t *s, size_t length, coap_option_num_t optnum
       default:
         /* add segment */
         optlist = coap_new_optlist(optnum, s - p, p);
+        if (!optlist) {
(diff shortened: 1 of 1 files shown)


mrdeep1:
These 3 commits will need to be merged into 1 at some point.  Otherwise, they look to be ok.

Oblivionsage:
Done! Squashed into a single commit. Thanks for the review.

mrdeep1:
Thanks for doing this.

PR #1860, opened 2026-01-05
feat(mbedtls): add support for PSA crypto interface

mahavirj:
- Changes to align with https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-4.0.0
- Pending - Enable mbedtls-4.0 build in CI

### DIFF
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -65,20 +65,59 @@
 #include <mbedtls/platform.h>
 #include <mbedtls/net_sockets.h>
 #include <mbedtls/ssl.h>
+#include <mbedtls/version.h>
+
+/* Auto-detect mbedTLS 4.x which requires PSA Crypto APIs */
+#if MBEDTLS_VERSION_NUMBER >= 0x04000000
+#define COAP_USE_PSA_CRYPTO 1
+#else
+#define COAP_USE_PSA_CRYPTO 0
+#endif
+
+#if COAP_USE_PSA_CRYPTO
+#include <psa/crypto.h>
+#else
 #include <mbedtls/entropy.h>
 #include <mbedtls/ctr_drbg.h>
+#include <mbedtls/sha256.h>
+#include <mbedtls/md.h>
+#include <mbedtls/cipher.h>
+#endif /* COAP_USE_PSA_CRYPTO */
 #include <mbedtls/error.h>
 #include <mbedtls/timing.h>
 #include <mbedtls/ssl_cookie.h>
 #include <mbedtls/oid.h>
 #include <mbedtls/debug.h>
-#include <mbedtls/sha256.h>
 #if defined(ESPIDF_VERSION) && defined(CONFIG_MBEDTLS_DEBUG)
 #include <mbedtls/esp_debug.h>
 #endif /* ESPIDF_VERSION && CONFIG_MBEDTLS_DEBUG */
-#if defined(MBEDTLS_PSA_CRYPTO_C)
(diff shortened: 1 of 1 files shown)


mrdeep1:
In principle, I support this type of change to work with later versions of a llibrary, and thanks for doing the work.

However, I would prefer the static inline functions in `coap_mbedtls_internal.h` to be in `coap_mbedtls.c` (not as inline) for the simplicity of line by line debugging.  This does then raise the question as to how much `coap_mbedtls_internal.h` is actually needed.

The 3 `switch ` compile failures can be fixed by doing something similar to `switch((int)alg) {` as you are reporting the missing case values.

mahavirj:
> In principle, I support this type of change to work with later versions of a llibrary, and thanks for doing the work.
> 
> However, I would prefer the static inline functions in `coap_mbedtls_internal.h` to be in `coap_mbedtls.c` (not as inline) for the simplicity of line by line debugging. This does then raise the question as to how much `coap_mbedtls_internal.h` is actually needed.
> 
> The 3 `switch ` compile failures can be fixed by doing something similar to `switch((int)alg) {` as you are reporting the missing case values.

Sure. Have merged in the abstraction in same file. Hopefully the CI passes this time around. Have verified the tests locally with both mbedtls-3.x and mbedtls-4.0 release.

mrdeep1:
Thanks for doing this.

In principal, it looks to be fine.  I will be running my local test suites for both v3.6.0 and v4.0.0 against this PR to check for any unexpected behaviours.  These suites run for hours.

Not sure how to add in v4.0.0 build testing into the CI build other than by cloning MbedTLS, building it and using it.

mrdeep1:
The MbedTLS v3.6.0 regression tests have worked as expected, v4.0.0 still has a few hours to run.

Several things have come to light which are fixed by 
[mbedtls_patch.txt](https://github.com/user-attachments/files/24468596/mbedtls_patch.txt)
 which fixes a memory leak caused by memory allocation by `psa_crypto_init()` which is not cleaned up at application exit, along with some code (SHA256) is not required for client only builds.

As FYI `coap_dtls_startup()` and  `coap_dtls_shutdown()` are invoked by `coap_startup()` and `coap_cleanup()` respectively.

mrdeep1:
When running with v4.0.0, for PSK there was a (server) core dump, and with PKI differences in behaviour when handling error conditions which need investigating.

mahavirj:
> When running with v4.0.0, for PSK there was a (server) core dump, and with PKI differences in behaviour when handling error conditions which need investigating.

Are these extended tests available in the repository, any instruction pointers? I can help to check the failure once.

mrdeep1:
> Are these extended tests available in the repository, any instruction pointers?

Unfortunately no, they are very platform / TLS library dependant and need careful analysis to understand what is happening when changes occur.

The PSK issue should be easy to fix - I suspect an Identity + DTLS 1.3 incompatibility.  For the PKI, it could be that previous versions are not correctly handling some certificate mismatch scenarios.

I will try to get this sorted out today.

mrdeep1:
[mbedtls_patch2.txt](https://github.com/user-attachments/files/24501117/mbedtls_patch2.txt) fixes all issues found so far and ready for for release with this patch.

For PKI, it was a v3.6.0 issue where `authmode `was not being correctly set for a client in `ssl_tls13_validate_certificate()` which then cause a "No CA Chain" issue for anonymous (no client certificates defined) clients. v3.6.0 usage has been fixed.

For PSK is was a premature call to `mbedtls_ssl_set_hs_psk()` in the interface into v4.0.0.

mrdeep1:
Thanks for doing this work.

PR #1861, opened 2026-01-08
Disable CSM for non-reliable proto

magdalenaszumny:
Disable Capabilities and Settings Messages for non-reliable protocols

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -1351,6 +1351,12 @@ coap_send_test_extended_token(coap_session_t *session) {
 int
 coap_client_delay_first(coap_session_t *session) {
 #if COAP_CLIENT_SUPPORT
+
+  /* Disable Capabilities and Settings Messages for non-reliable protocols */
+  if (COAP_PROTO_NOT_RELIABLE(session->proto)) {
+    session->doing_first = 0;
+  }
+
   if (session->type == COAP_SESSION_TYPE_CLIENT && session->doing_first) {
     int timeout_ms = 5000;
     coap_session_state_t current_state = session->state;


mrdeep1:
This fix will have a lot of unintended consequences.  `doing_first `is set whenever a session start-up sequence has to be done in a particular order - i.e. a start-up response is required before the next request is transmitted. `coap_client_delay_first()` enforces that the start-up sequence has been completed (or times out where things get handled appropriately).

It is set for CSM, Extended Token support, Q-Block support and OSCORE.  The latter 3 can all be using unreliable protocols.

A lot of work was done in #1833 to avoid a single thread having to wat on `doing_first` and it possible something got broken here.

It would be good (as an attachment) to get a copy of the debug enabled log file to demonstrate the issue that you are having (i.e. not not including this PR) so that the real problem can get fixed.

mrdeep1:
@magdalenaszumny It would be helpful to have feedback, so this issue can be properly resolved.

magdalenaszumny:
Yes, of course! I asked my colleague to add some more explanation here as he was more involved in this issue. I will remind him

wwiercigroch:
For us, the main problem is the way these parameters are negotiated, which blocks the client for long periods of time in various places in the API. We use io with events, and completely blocking the client thread for 5 seconds is a big problem for us, especially since we don't use these features. While this can be bypassed in the case of Q-Block and Extended Token (I wasn't familiar with OSCORE), sending the first packet always blocks due to CSM, despite using UDP. However, this proposal isn't the best one due to using `coap_client_delay_first` to handle all cases. I also see changes to `coap_send` (we have an older version), which was one of the causes of the problem. I think we can close this thread for now. If CSM continues to be an issue, we'll consider and propose a different solution.

mrdeep1:
Thanks for the update. 

> sending the first packet always blocks due to CSM, despite using UDP

 With release 4.3.5 I cannot find where `session->doing_first` (and hence wait in `coap_client_delay_first()`) is set for a non-reliable protocol - e.g. UDP or DTLS other than for Extended Token, Q-Block or OSCORE, so am surprised there is any notion of blocking on a CSM exchange for a non-reliable protocol.

The documentation for `coap_add_data_large_request_lkd()` or` coap_session_max_pdu_size_lkd()` says
````
  /*
   * Delay if session->doing_first is set.
   * E.g. Reliable and CSM not in yet for checking block support
   */
````
Which may be confusing you about the CSM.

The CSM check (setting of `session->doing_first`) is done in `coap_session_check_connect()`.

What I think is likely to be happening is that you are calling` coap_context_set_block_mode() `with a parameter value that includes COAP_BLOCK_TRY_Q_BLOCK (value 0x40) set.

mrdeep1:
Closing as later (post 4.3.5) #1833 code reduces the requirement to wait on functional support protocol exchanges to take place before any application data can be transmitted.

PR #1865, opened 2026-01-21
oss-fuzz: Add new fuzzer for oscore pdu decryption

arthurscchan:
This PR introduces a new fuzzer for libcoap's OSCORE decryption functionality, covering OSCORE-protected PDU parsing, cryptographic decryption operations, and security context handling. The fuzzer adopts a simplified structure consistent with existing libcoap fuzzers, targeting OSCORE decryption logic with randomised context.

Remark: This is the first in a series of PRs intended to expand fuzzing coverage for the LibCoAP project. Subsequent PRs will continue to address further coverage gaps and enhance fuzzing effectiveness, thereby improving the detection of issues across the codebase.

### DIFF
--- a/tests/oss-fuzz/oscore_decrypt_target.c
+++ b/tests/oss-fuzz/oscore_decrypt_target.c
@@ -1,0 +1,100 @@
+#include "coap3/coap_internal.h"
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  coap_context_t *ctx;
+  coap_session_t *session = NULL;
+  coap_pdu_t *pdu = NULL, *decrypted = NULL;
+  coap_address_t addr;
+  size_t opt_len, encrypted_len;
+  uint16_t partial_iv_len, kid_context_len, kid_len;
+  uint8_t oscore_option[256];
+  static const char oscore_conf_str[] =
+      "master_secret,hex,\"0102030405060708090a0b0c0d0e0f10\"\n"
+      "master_salt,hex,\"9e7ca92223786340\"\n"
+      "sender_id,hex,\"\"\n"
+      "rfc8613_b_1_2,bool,false\n"
+      "recipient_id,hex,\"01\"\n";
+
+  if (size < 2)
+    return 0;
+
+  /* Initialize libcoap with OSCORE support */
+  coap_startup();
+  coap_set_log_level(COAP_LOG_ERR);
+  coap_str_const_t conf = { sizeof(oscore_conf_str) - 1, (const uint8_t *)oscore_conf_str };
+  coap_oscore_conf_t *oscore_conf = coap_new_oscore_conf(conf, NULL, NULL, 0);
+
+  /* Stop transmission of any error packets */
+  coap_debug_set_packet_loss("1-10");
+
+  /* Set up OSCORE context */
+  ctx = coap_new_context(NULL);
+  if (ctx && oscore_conf) {
+    if (coap_context_oscore_server(ctx, oscore_conf)) {
+      /* Create a new client session */
+      coap_address_init(&addr);
+      addr.size = sizeof(struct sockaddr_in);
+      addr.addr.sin.sin_family = AF_INET;
+      addr.addr.sin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addr.addr.sin.sin_port = htons(5683);
+
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Build PDU with OSCORE option and encrypted data from fuzzer data */
+      if (session) {
+        /* Calculate OSCORE option length based on RFC 8613 Section 6.1 */
+        partial_iv_len = data[0] & 0x7;
+        if (1UL + partial_iv_len > size)
+          goto cleanup;
+
+        kid_context_len = data[0] & 0x10 ? 1 + data[1 + partial_iv_len] : 0;
+        /* 1s are byte 0 and kid length to be added */
+        if (1UL + partial_iv_len + kid_context_len + 1 > size)
+          goto cleanup;
+
+        /* Build option with KID forced to one byte to match recipient_id in oscore_conf_str */
+        kid_len = 1;
+        opt_len = 1 + partial_iv_len + kid_context_len + kid_len;
+        memcpy(oscore_option, data, opt_len - 1);
+        oscore_option[opt_len - 1] = 0x01;
+
+        pdu = coap_pdu_init(COAP_MESSAGE_NON, COAP_REQUEST_CODE_GET,
+                            coap_new_message_id(session), 256);
+
+        if (pdu) {
+          /* Add OSCORE option with KID fixed to 0x01 */
+          if (coap_add_option(pdu, COAP_OPTION_OSCORE, opt_len, oscore_option)) {
+            /* Add remaining data as encrypted payload (from after original option) */
+            encrypted_len = size - opt_len;
+            if (encrypted_len > 0) {
+              coap_add_data(pdu, encrypted_len, &data[opt_len]);
+            }
+
+            /* Test OSCORE decryption */
+            decrypted = coap_oscore_decrypt_pdu(session, pdu);
+            if (decrypted) {
+              coap_show_pdu(COAP_LOG_ERR, decrypted);
+              coap_delete_pdu(decrypted);
+            }
+          }
+          coap_delete_pdu(pdu);
+        }
+cleanup:
+        coap_session_release(session);
+      }
+    } else {
+      /* coap_context_oscore_server failed, oscore_conf not consumed */
+      coap_delete_oscore_conf(oscore_conf);
+    }
+  } else if (oscore_conf) {
+    /* ctx creation failed, need to free oscore_conf */
+    coap_delete_oscore_conf(oscore_conf);
+  }
+
+  if (ctx)
+    coap_free_context(ctx);
+
+  coap_cleanup();
+  return 0;
+}


mrdeep1:
Thanks for doing this work.  The more the testing of the libcoap code, the better.

There are some compilation warnings that need to be addressed.  If (in the libcoap directory), you do `make -C tests/oss-fuzz -f Makefile.oss-fuzz check clean` following a `./configure` and correct the compilation warnings.

The standard for libcoap coding documentation is to use `/* ... */`, not `// ..`. Please update accordingly.

mrdeep1:
If would help is this PR can be rebased against the latest develop code so that any oss-fuzz compilation errors can be picked up before the next push.

arthurscchan:
@mrdeep1 I have fixed them accordingly.

mrdeep1:
Your external recent changes to oss-fuzz projects/libcoap/build.sh result in
````
configure: WARNING: ==> Using the configure options '--with-gnutls', '--with-openssl', '--with_wolfssl', '--with-mbedtls' or '--with-tinydtls' without '--enable-dtls' is useless and will be ignored.

from

./autogen.sh && ./configure --disable-doxygen --disable-manpages \
                            --disable-dtls --enable-tests        \
                            --with-openssl --enable-oscore       \
                            --enable-server                      \
````
As I read your test fuzz
1. The first 36 bytes are configuration data.  It is not likely that the configuration data will successfully create an OSCORE configuration file - `coap_new_oscore_conf()` is worthy of its own fuzzer.
2. If not fail, then next 4 bytes are common to the recipient ID and the start of the PDU data - which is fine.
3. If not fail, then `coap_new_client_session()` will always fail as client code is not enabled in `build.sh` (if the correct option -`-enable-server-mode` is used).
4. If not fail, then `coap_parse_pdu()` is most likely to fail (but does have its own fuzzer).
5. If not fail, only then is `coap_oscore_decrypt_pdu()` actually fuzz tested.

I do not know enough about oss-fuzz to persuade it to use the the current pull request (if obgm/libcoap) or pushed branch in a fork instead of the develop branch.  All this needs to be properly tested before merging anything into the `develop` branch.

**Comment by mrdeep1** [comment]
The fuzzer is now run when you push a PR (after rebasing with develop).  I am considering restricting this to branch names that start with oss.

If you look at the output from the CI fuzzer, you now get
````
configure: WARNING: ==> --enable-oscore requires crypto support from TLS library or OS
````
after your latest changes to `oss-fuzz projects/libcoap/build.sh`.  This means that any encode / decode OSCORE function will fail.

`--enable-server` is an unrecognized option for configure. `--enable-server-mode` is the correct option, but using this without `--enable-client-mode` will disable any client code support.

I think `build.sh`' `./configure` parameters should read (server and client code are enabled by default, OSCORE core is enabled by default)
````
./autogen.sh && ./configure --disable-doxygen --disable-manpages \
                            --with-openssl --enable-tests     \
  && make -j$(nproc)   
````

**Comment by mrdeep1** [comment]
oss-fuzz has now been updated to be able to do OSCORE fuzzing with OpenSSL as the TLS encrypt / decrypt engine by https://github.com/google/oss-fuzz/pull/14845. 

A rebase with develop is needed before the next PR push.

**Comment by arthurscchan** [comment]
@mrdeep1 Thanks for the update, I will follow up on that. Sorry for the delay reply.

**Comment by mrdeep1** [comment]
I'm looking into the zephyr build issue and will get a fix done.

**Comment by arthurscchan** [comment]
@mrdeep1 Thanks. It seems to be a warning in the configuration of zephyr build on mbedlts module, not sure if it related to this. https://github.com/obgm/libcoap/pull/1805

**Comment by mrdeep1** [comment]
Rebase against `develop`, and the Zephyr issue is fixed, as well as a otential memory leak when parsing an OSCORE configuration file.  #1805 takes the Zephyr' MbedTLS configuration further.

With your fuzzer, it is unclear to me as to how you you seed the fuzzing data.

**Comment by AdamKorcz** [comment]
> With your fuzzer, it is unclear to me as to how you you seed the fuzzing data.

Currently, the fuzzer does not have a seed, but it will be able to explore its target when running on OSS-Fuzz; The coverage feedback will help the fuzzer generate meaningful testcases that make it progress. A dictionary might be helpful here and perhaps a few well-crafted seeds, but I think a good solution is also to let it run on OSS-Fuzz and observe how it progresses.

**Comment by arthurscchan** [comment]
@mrdeep1 Thanks for the update and the fixes. I have rebased the PR already. Once the CI is done, I think it is good to go on my side.

**Comment by mrdeep1** [comment]
CiFuzz is now running as a part of the CI.

**Comment by mrdeep1** [comment]
> Currently, the fuzzer does not have a seed, but it will be able to explore its target when running on OSS-Fuzz;
> The coverage feedback will help the fuzzer generate meaningful testcases that make it progress. A dictionary
> might be helpful here and perhaps a few well-crafted seeds, but I think a good solution is also to let it run
> on OSS-Fuzz and observe how it progresses.

I don't see any` return -1; ` in `LLVMFuzzerTestOneInput()` to control what goes into the corpus.

**Comment by arthurscchan** [comment]
@mrdeep1 For OSS-Fuzz execution, it leverages libFuzzer (or other engine) for coverage-guided fuzzing. When a test input triggers new code coverage or explores unique execution paths, libFuzzer (or other engine) automatically saves that input as part of its corpus. OSS-Fuzz does not use the return value of LLVMFuzzerTestOneInput to determine if a test case is interesting or should be added to the corpus. Instead, it relies on code coverage and crash detection directly.

**Comment by mrdeep1** [comment]
The current CiFuzz run output reports
````
INFO: A corpus is not provided, starting from an empty corpus
#2	INITED exec/s: 0 rss: 84Mb
WARNING: no interesting inputs were found so far. Is the code instrumented for coverage?
This may also happen if the target rejected all inputs we tried so far
````
I suspect that the initial `coap_new_oscore_conf()` always fails, so none of the other code gets executed.

File: tests/oss-fuzz/oscore_decrypt_target.c [hunk 1]
```diff
@@ -0,0 +1,70 @@
...
+
+  if (size < 38)
+    return 0;
+
+  /* Initialize libcoap with OSCORE support */
+  coap_startup();
+  coap_set_log_level(COAP_LOG_EMERG);
+  memcpy(&conf_buf, data, 36);
+  coap_str_const_t conf = { 36, conf_buf };
+  coap_oscore_conf_t *oscore_conf = coap_new_oscore_conf(conf, NULL, NULL, 0);
```
mrdeep1, line 26: Unless the OSCORE configuration file is well formed, `oscore_conf` will end up as NULL when doing `coap_new_oscore_conf()`.  The first 36 random bytes will not work well here.

My suggestion is that you separately fuzz `coap_new_oscore_conf()`, and use a working configuration file so that the following code can be properly tested.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,70 @@
...
+      /* coap_context_oscore_server takes ownership of oscore_conf */
+      oscore_conf = NULL;
+
+      /* Add recipient with ID from fuzzer data */
+      coap_bin_const_t recipient_id = { 4, &data[36] };
+      coap_new_oscore_recipient(ctx, &recipient_id);
+
+      /* Create a new client session */
+      coap_address_init(&addr);
+      addr.size = sizeof(struct sockaddr_in);
```
mrdeep1, line 37: addr.size is set in `coap_address_init()`, so not needed here.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,70 @@
...
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Parse and attempt to decrypt the PDU from fuzzer data */
+      if (session) {
+        data += 37;
+        size -= 37;
+        pdu = coap_pdu_init(0, 0, 0, size);
+        if (pdu && coap_pdu_parse(COAP_PROTO_UDP, data, size, pdu)) {
+          decrypted = coap_oscore_decrypt_pdu(session, pdu);
```
mrdeep1: If the resultant parsed PDU does not contain an OSCORE option, there currently is an `assert()` if this option is missing right at the start of `coap_oscore_decrypt_pdu()`.  Either the `assert()` needs to be removed or a check for the OSCORE option.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Thanks for the suggestion. I will fix them accordingly.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Thanks for the suggestion. I will fix them accordingly.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: I think the better way is to use static configuration and concentrate on the decrypt first, and I will leave the configuration fuzzing to later PRs.

mrdeep1: > I think the better way is to use static configuration and concentrate on the decrypt first


Works for me.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,62 @@
...
+  /* Initialize libcoap with OSCORE support */
+  coap_startup();
+  coap_set_log_level(COAP_LOG_EMERG);
+  coap_str_const_t conf = { sizeof(oscore_conf_str) - 1, (const uint8_t *)oscore_conf_str };
+  coap_oscore_conf_t *oscore_conf = coap_new_oscore_conf(conf, NULL, NULL, 0);
+
+  /* Set up OSCORE context */
+  ctx = coap_new_context(NULL);
+  if (ctx && oscore_conf) {
+    if (coap_context_oscore_server(ctx, oscore_conf)) {
```
mrdeep1, line 34: If `ctx == NUL`L, `oscore_conf `never gets freed off (memory leak).

tests/oss-fuzz/oscore_decrypt_target.c
mrdeep1: setting `addr.size` is not needed (and is not supported for some OS builds).  Differences in `addr `are taken care of in `coap_address_init()` where `addr.size` is set if appropriate.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Thanks, I've fixed that. I've also fixed the fuzzer by adding some additional checks to avoid entering the error label path that requires an ACK, which would otherwise crash the fuzzer in the assert statement.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,77 @@
...
+      addr.addr.sin.sin_port = htons(5683);
+
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Parse and attempt to decrypt the PDU from fuzzer data */
+      if (session) {
+        pdu = coap_pdu_init(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET, 0, size);
+
+        /* Only proceed if parse succeeds without warnings */
+        if (pdu && coap_pdu_parse(COAP_PROTO_UDP, data, size, pdu) == size) {
```
mrdeep1: coap_pdu_parse() only returns 0 or 1.  See [coap_pdu_parse](https://libcoap.net/doc/reference/develop/group__pdu.html#gab9981ea277da8bc4e5c4f82d6ca4c006)

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,77 @@
...
+          coap_pdu_set_code(pdu, COAP_REQUEST_CODE_GET);
+
+          coap_opt_iterator_t opt_iter;
+          coap_opt_t *option = coap_check_option(pdu, COAP_OPTION_OSCORE, &opt_iter);
+
+          if (option) {
+            /* Skip if OSCORE option length is invalid */
+            size_t oscore_len = coap_opt_length(option);
+            if (oscore_len >= 1 && oscore_len <= 50) {
```
mrdeep1: Not sure why you need this as it restricts fuzzing `coap_oscore_decrypt_pdu()`.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,77 @@
...
+      if (session) {
+        pdu = coap_pdu_init(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET, 0, size);
+
+        /* Only proceed if parse succeeds without warnings */
+        if (pdu && coap_pdu_parse(COAP_PROTO_UDP, data, size, pdu) == size) {
+          coap_pdu_set_code(pdu, COAP_REQUEST_CODE_GET);
+
+          coap_opt_iterator_t opt_iter;
+          coap_opt_t *option = coap_check_option(pdu, COAP_OPTION_OSCORE, &opt_iter);
+
```
mrdeep1, line 64: Consider using `coap_insert_option()` (internal function) to add in an OSCORE option if it does not exist.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,78 @@
...
+      addr.size = sizeof(struct sockaddr_in);
+      addr.addr.sin.sin_family = AF_INET;
+      addr.addr.sin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addr.addr.sin.sin_port = htons(5683);
+
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Parse and attempt to decrypt the PDU from fuzzer data */
+      if (session) {
+        pdu = coap_pdu_init(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET, 0, size);
```
mrdeep1: If you are going to add in an OSCORE option, this will fail as the PDU cannot expand. The max size parameter (currently `size`), needs to be at least `2*size +3`.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,78 @@
...
+        if (pdu && coap_pdu_parse(COAP_PROTO_UDP, data, size, pdu)) {
+          coap_pdu_set_code(pdu, COAP_REQUEST_CODE_GET);
+
+          coap_opt_iterator_t opt_iter;
+          coap_opt_t *option = coap_check_option(pdu, COAP_OPTION_OSCORE, &opt_iter);
+
+          /* Add OSCORE option with fuzzer data if not present */
+          if (!option && size > 0) {
+            coap_insert_option(pdu, COAP_OPTION_OSCORE, size, data);
+          }
```
mrdeep1: This `coap_insert_option()` will fail if size > 255 bytes, and will also fail if the PDU cannot be expanded beyond the defined max size.

I suggest that you re-check that the OSCORE option is now present.

`coap_insert_option()` does return 0 on failure.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Hi, I remove the use of random data to generate the options again. Instead I use a single byte static option flag. Thus I think the size is good enough to set at size + 8, which leave enough space for the encoding overhead and the static options if it needed to be added.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Fixed.

arthurscchan:
@mrdeep1 
At this point, I have found that the code actually crashes due to a coap_lock_check, because some of the paths for the PDU decryption functions fail and go to the error label, which then sends an ACK under thread-safe mode with a locking mechanism. I have spent some time trying to disable or manually acquire the lock (by `coap_lock_lock` and `coap_lock_unlock` macro combo) before calling the decrypt function, but I have not managed to make it work. I am considering disabling the thread-safe settings completely for fuzzing. As in my understanding, OSS-Fuzz is supposingly running in single thread. Will adding the `--disable-thread-safe` flag to the configure script have a significant effect on other fuzzers and the overall fuzzing build? I would like to confirm this before making changes in OSS-Fuzz. Thanks for the great effort to support me on creating the fuzzer and sorry to bother you that much because of my unfamiliarity of the project.

For reference, the failing in the CIFuzz is actually the thread-safe locking problem, which I tested in my local environment if I disable the thread-safe completely, the fuzzer will work.

mrdeep1:
--disable-thread-safe configure option should be fine.  I will push a PR now for that to oss-fuzz.

mrdeep1:
Re-run with new oss-fuzz environment.

To stop sending an ACK, it is worth trying to make the PDU type NON, not CON.

arthurscchan:
@mrdeep1 Thanks. I have tried both the NON and CON types before, and although other parts work fine for NON, the default path in the switch statement for option checking keeps failing and jumps to the goto label. This causes the ACK to be sent anyway. I assume this is because the option byte, parsed from random data, is invalid.

mrdeep1:
I've been thinking how to best exercise `coap_oscore_decrypt_pdu()`. An encrypted OSCORE PDU looks like
````
<header><optional token><OSCORE option><optional options><encrypted data (of original token + options + data)>
````
so OSCORE option and encrypted data only need to be fuzzed to test `coap_oscore_decrypt_pdu()`.  So, if you
````
coap_pdu_init()                      // With space to hold OSCORE option and data
coap_add_option(OSCORE)
coap_add_data()                     // With fuzz data
coap_oscore_decrypt_pdu()
````
If the OSCORE option is randomized (other a zero length option, or a 1 byte option with a zero value), then based on byte 0 value (and possibly byte size s) the overall length needs to be correct based on kid, kid context and Partial IV.  See [RFC 8613 6.1](https://datatracker.ietf.org/doc/html/rfc8613#section-6.1).

**Comment by mrdeep1** [comment]
> This causes the ACK to be sent anyway.

Actually No. As per `coap_send_ack_lkd()`, it will only get sent if CON and unreliable (i.e. UDP or DTLS) protocol.

**Comment by arthurscchan** [comment]
Thank you for your explanation and suggestions.

My understanding was that, by parsing random data into a PDU, we could avoid the need for precise length calculations, as the parser would handle the structure for us. I wanted to confirm whether the intended flow you mentioned, initialisation, adding the OSCORE option, adding data, and then decrypting, is specifically designed to restrict the fuzzed input to just the encrypted data, in order to focus on testing the behaviour of the decrypt function.

However, I was considering whether, that approach maybe result in a strict mapping of pdu fields and the encrypted data. If the parsing succeeds but the resulting PDU contains fields that do not correspond to valid encrypted data, this might actually allow the fuzzer to explore a wider range of code paths, including error handling scenarios. Would it be beneficial to allow the fuzzer to generate such cases, in order to increase coverage of potential error paths?

Given that the purpose of fuzzing is to explore a wide range of code paths, including those that may trigger error handling, I wonder if there might be value in allowing the fuzzer to generate cases where the PDU structure is valid but the contents are not consistent with a properly encrypted message even if we are only targeting the oscore_decrypt_pdu function. I would appreciate your thoughts on whether it is preferable to constrain the input strictly to valid OSCORE option and data fields, or if it would be beneficial to also explore these error scenarios. Your guidance on this would be most helpful. Thanks a lot in advance for your help.

**Comment by mrdeep1** [comment]
Yes, parsing the provided data does all the necessary syntax checks when building up the resultant PDU (or has a fail return).  What we end up with valid provided data is
````
<header><optional token><optional options><optional data>
````
and for OSCORE, there needs to be an OSCORE CoAP option and there needs to be encrypted data if `coap_oscore_decrypt_pdu()` has any chance of being successful.

While `coap_parse_pdu()` will check the length of an OSCORE option (it can be 0 - 255 bytes long), it does not check the internal syntax of the option - `coap_oscore_decrypt_pdu()` internally does that using `oscore_decode_option_value()`.  So if the syntax is messed up, most of the `coap_oscore_decrypt_pdu()` will not get exercised. Hence my comment that you need to compute the length of the OSCORE option (when adding it to the PDU) based on the its fuzz data so that the code can get past `oscore_decode_option_value()` with some bad values derived from the OSCORE option.

So in answer to your first question - I want the `coap_oscore_decrypt_pdu()` exercised as much as possible by including random values in both the OSCORE option and encrypted data, not just in the encrypted data.

Question 2.  There will be a lot of random data, with a high chance (if able to decrypt the encrypted data) that the resultant original unencrypted PDU will be broken and not parse correctly.

Question 3. I think with my approach `coap_oscore_decrypt_pdu()`  will get well exercised.

**Comment by arthurscchan** [comment]
@mrdeep1 Thank you for your detailed explanation. I am now fixing the fuzzers accordingly and doing some local testing with debug logging to help me confirm and test that the new approach works and stretch through the coap_oscore_decrypt_pdu function as much as possible. I will give you an update and have the updated fuzzers in this PR after I fix and complete my local testing. Thanks.

**Comment by mrdeep1** [comment]
It is worth making the change done in #1881 to remove the need to declare internal functions (which may change over time as they are not a part of the public API).

File: tests/oss-fuzz/oscore_decrypt_target.c [hunk 10]
```diff
@@ -0,0 +1,84 @@
...
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Build PDU with OSCORE option and encrypted data from fuzzer data */
+      if (session) {
+        pdu = coap_pdu_init(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET,
+                            coap_new_message_id(session), 256);
+
+        if (pdu) {
+          /* OSCORE option length from first byte (0-255) */
+          opt_len = data[0];
```
mrdeep1: If you read the definition of the OSCORE option (https://datatracker.ietf.org/doc/html/rfc8613#section-6.1), this actually is

````
partial_iv_len = data[0] & 0x7;
kid_context_len = (data[0] & 0x10) ? data[partial_iv_len+1] + 1: 0;
/* kid len has to be a min of 1 if k bit set - can be randomly larger */
kid_len = (data[0] & 0x8) ? 1 : 0;
opt_len = 1 + partial_iv_len  + kid_context_len  + kid_len
```` 
Note that `opt_len` cannot be greater than `size `or greater that 255 (as would be picked up by `coap_pdu_parse()` on receipt of a PDU).

File: tests/oss-fuzz/oscore_decrypt_target.c [hunk 11]
```diff
@@ -0,0 +1,84 @@
+#include <coap3/coap.h>
+#include <arpa/inet.h>
+
+/* Declare internal prototypes */
+extern coap_pdu_t *coap_oscore_decrypt_pdu(coap_session_t *session, coap_pdu_t *pdu);
```
mrdeep1: Replace with `#include "coap3/coap_internal.h"` as is done for the other fuzzers so there is no longer any need to define internal functions (which may change syntax over time).

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,84 @@
...
+      addr.size = sizeof(struct sockaddr_in);
+      addr.addr.sin.sin_family = AF_INET;
+      addr.addr.sin.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+      addr.addr.sin.sin_port = htons(5683);
+
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Build PDU with OSCORE option and encrypted data from fuzzer data */
+      if (session) {
+        pdu = coap_pdu_init(COAP_MESSAGE_CON, COAP_REQUEST_CODE_GET,
```
mrdeep1: Make this COAP_MESSAGE_NON to stop error PDUs getting transmitted.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,84 @@
...
+                            coap_new_message_id(session), 256);
+
+        if (pdu) {
+          /* OSCORE option length from first byte (0-255) */
+          opt_len = data[0];
+          if (opt_len > size - 1) opt_len = size - 1;
+
+          /* Add OSCORE option with random fuzzer data (can be 0-length) */
+          coap_add_option(pdu, COAP_OPTION_OSCORE, opt_len, opt_len > 0 ? &data[1] : NULL);
```
mrdeep1: Should read (f you are using the first opt_len bytes to create the random, but valid syntax OSCORE option)
````
coap_add_option(pdu, COAP_OPTION_OSCORE, opt_len, data);
````

File: tests/oss-fuzz/oscore_decrypt_target.c [hunk 14]
```diff
@@ -0,0 +1,84 @@
...
+          /* Add remaining data as encrypted payload */
+          encrypted_len = size - 1 - opt_len;
+          if (encrypted_len > 0) {
+            coap_add_data(pdu, encrypted_len, &data[1 + opt_len]);
+          }
+
+          /* Verify OSCORE option exists before attempting decryption */
+          coap_opt_iterator_t opt_iter;
+          if (coap_check_option(pdu, COAP_OPTION_OSCORE, &opt_iter)) {
+            /* Test OSCORE decryption */
```
mrdeep1: No need to do this as you have added the OSCORE option.

mrdeep1 (changes requested):

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Have fixed accordingly.

arthurscchan: I have also added a second checking to make sure data does enough for the kid_context_len before doing the assignment.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: fixed.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Fixed, and rebased with that.

tests/oss-fuzz/oscore_decrypt_target.c
arthurscchan: Yes, sorry for not noticing that.

mrdeep1:
I suggest you make the following changes.  
[oscore.patch.txt](https://github.com/user-attachments/files/25113964/oscore.patch.txt)

`rfc8613_b_1_2 `set to false disables the initial 4.01 response before doing any decoding.

mrdeep1:
As error packets can get transmitted, it is worth disabling any packet transmission using `coap_debug_set_packet_loss()` as in
[oscore.patch1.txt](https://github.com/user-attachments/files/25123731/oscore.patch1.txt).
 If you want to see what is happening, I suggest you (temporarily) set the logging level to COAP_LOG_DEBUG or possibly COAP_LOG_OSCORE.

mrdeep1:
Apologies for the _loss/_lost typo.

tests/oss-fuzz/oscore_decrypt_target.c
```diff
@@ -0,0 +1,95 @@
...
+      addr.addr.sin.sin_port = htons(5683);
+
+      session = coap_new_client_session(ctx, NULL, &addr, COAP_PROTO_UDP);
+
+      /* Build PDU with OSCORE option and encrypted data from fuzzer data */
+      if (session) {
+        /* Calculate OSCORE option length based on RFC 8613 Section 6.1 */
+        partial_iv_len = data[0] & 0x7;
+        if (1UL + partial_iv_len > size)
+          return 0;
```
mrdeep1: All these return 0 need to go to go to a cleanup statement so all mallocs are freed off.

arthurscchan: Yes, I just see that, I have fixed, now doing some final testing.

mrdeep1:
To fix a lot of the `OSCORE: Security Context not found` reported issues, the KID value in the OSCORE option needs to be set to `recipient_id `value in `oscore_conf_str`.

arthurscchan:
I have changed the logic a bit, by still using the flags and random option but fixed the kid to be 0x01 to match the receipient ID in the conf_str.

mrdeep1:
I would like to see kid_context randomized as this causes different paths to be taken in the OSCORE PDU decode logic.

See 
[oscore.patch2.txt](https://github.com/user-attachments/files/25258430/oscore.patch2.txt) as a suggestion.

At some point all the commits need to b squashed into a single commit.

arthurscchan:
I have applied the fix to allow KID randomised, and I have Also rebase the commits.

mrdeep1:
Thanks for this.  The 2 commits need to be squashed into a single commit, then this PR is good to be merged.

This randomizing of `kid_context `is generating some `CRIT OSCORE: Security Context not found,` but this is fine.

PR #1878, opened 2026-01-30
remove unnecessary build checks

casaroli:
These macros are not used anywhere in Zephyr, and check_include_file is slow, so these are unnecessary.

This should make the configuration step of the build (cmake) faster and should not break anything.


casaroli:
I think there is a very small chance that Riot, Contiki or Windows might even require these, but I believe this would be a problem with their build system.

mrdeep1:
I have tidied up the required #include system header files in #1879.  `coap_internal.h` is now where all the header HAVE_ definitions are now used.

This will affect the no longer needed checks in this PR.

PR #1884, opened 2026-02-13
OSS-Fuzz: Add oscore conf parse fuzzer

arthurscchan:
This PR follows up on #1865 and creates a separate fuzzer for oscore conf parsing and option decoding.


mrdeep1:
These 2 commits need to be squashed into 1.  I suggest in the future that you just do `git rebase develop` with the develop branch updated to the latest, rather than create a separate commit.

arthurscchan:
Sorry and thanks for the suggestion. I have rebased on the develop branch.

PR #1886, opened 2026-02-16
Add sorting of optlist

magdalenaszumny:
Add sorting of optlist functionality to API.


mrdeep1:
I'm not sure why you need to do this, given that `coap_add_optlist_pdu()` does the sort for you.  Further more, man page documentation needs to be added for other users.

magdalenaszumny:
Thank you for taking a look! We are adding on layer above libcoap some abstraction layer so the user has easy interface and we are storing optlist as part of identifying the connection. It is easier for us to have it sorted for later comparison. 
I will add documentation

mrdeep1:
Thanks for this - looks fine to me.

PR #1891, opened 2026-02-17
Option to disable dtls

magdalenaszumny:


mrdeep1:
Not sure why you need to disable DTLS within coap_mbedtls.c as that is done by not having COAP_WITH_LIBMBEDTLS set.

magdalenaszumny:
> Not sure why you need to disable DTLS within coap_mbedtls.c as that is done by not having COAP_WITH_LIBMBEDTLS set.

I would like to use psa crypto from mbedtls but I would not like to use DTLS. My mbedtls configuration does not include functions for dtls, so I would like to have possibility in coap_mbedtls.c to not compile code related to dtls

magdalenaszumny:
I will close this PR, because it is getting more and more complicated :) For time being I will patch it on my side and maybe in future I will have some time to think about different solution that would not require so many $ifdefs, but would allow to use psa crypto from mbedtls without dtls.

mrdeep1:
When you say disable DTLS, do you mean disable both DTLS (ie UDP support) and TLS (ie TCP support) or just DTLS.  Support for disabling UDP would be a large change.

qavaxb:
Hi,
I think (D)TLS is not the only way of securing transport over CoAP. OSCORE is another way.
To support OSCORE, additional functions supporting cryptographic operations are required. For embedded targets using Zephyr, PSA Crypto API is preferred API. MbedTLS is supporting PSA Crypto API, but it is also library enabling TLS support.

So to use OSCORE with PSA Crypto API, but without (D)TLS, some additional conditions shall be made to the code.

mrdeep1:
See #2049 for a solution for this.

qavaxb:
Thank you for taking care of it.
I will try to test it on our side.

PR #1897, opened 2026-02-25
OSS-Fuzz: Add fuzzer for coap_net targets

arthurscchan:
This PR creates a new fuzzer targeting functions in coap_net source code.


mrdeep1:
Thanks.

PR #1898, opened 2026-02-25
Enhance configuration for zephyr

qavaxb:
I would like to enhance configuration from Zephyr Kconfig perspective.

Problematic though might be changing internal configuration. I would gladly accept suggestions.


mrdeep1:
In general, I have no issues with this enhancement.  A couple of the options are also configurable at run time (`coap_context_set_session_timeout()` and `coap_session_set_mtu()`).

COAP_DEFAULT_MAX_PDU_RX_SIZE usually ends up being larger (by default) if TCP is supported and is actually not associated with block size - see [RFC8323 5.3.1](https://datatracker.ietf.org/doc/html/rfc8323#section-5.3.1) second paragraph. However, this max value is different for other embedded O/S and it may make sense for the default for Zephyr to be different.

These 3 commits would need to be merged into 1 at some point.

qavaxb:
Thank you for comments. I have tried to get the same initial default values to minimize differences.
I'm aware that some of configurations are possible to change dynamically, however it is more convenient to have default value configured to most used in project.

I have aligned configuration names, so that it will be easier to track them - **CONFIG_LIB**COAP_DEFAULT_SESSION_TIMEOUT.

One commit is present.

mrdeep1:
Thanks

PR #1900, opened 2026-03-04
OSS-Fuzz: Fix Makefile for fuzzer to include CFLAGS from OSS-Fuzz

arthurscchan:
This PR fixes the Makefile for OSS-Fuzz fuzzers to preserve the CFLAGS from OSS-Fuzz which includes coverage instrumentation flags to allow coverage tracing for fuzzer code.


mrdeep1:
LGTM

PR #1901, opened 2026-03-04
OSS-Fuzz: Fix bias condition

arthurscchan:
This PR fixes a minor bias condition for one of the testing branch for fuzzer targeting coap_net.


mrdeep1:
LGTM

PR #1903, opened 2026-03-04
OSS-Fuzz: Simplify block_fuzzer

arthurscchan:
This PR simplify the block fuzzer and add in additional targets in the coap_block source code.


mrdeep1:
LGTM

PR #1908, opened 2026-03-06
OSS-Fuzz: Add more targets to network fuzzer

arthurscchan:
This PR expands the network_message_fuzzer to covers more targets in coap_net.

### DIFF
--- a/tests/oss-fuzz/network_message_target.c
+++ b/tests/oss-fuzz/network_message_target.c
@@ -2,79 +2,154 @@
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


mrdeep1:
These 2 commits need to be merged into 1.

arthurscchan:
sorry, my bad. Done now.

mrdeep1:
As you are "transmitting" PDUs, you need to make use of `coap_debug_set_packet_loss()` so that you do not create unnecessary network traffic when running oss-fuzz.

mrdeep1:
I think you need to rebase against the latest develop to fix the RIOT issue with network_message_target.c.

PR #1914, opened 2026-03-12
OSS-Fuzz: Add observe fuzzer

arthurscchan:
This PR creates a new fuzzer targeting pdu observer implementation.


mrdeep1:
Thanks for doing this work.

PR #1915, opened 2026-03-13
impr(security): tracking CVE using CPE

qavaxb:
As security awareness is increasing, I would like to propose changes that will help with tracking CVE within Zephyr environment.

This however is also connected with release process, as CPE reference contains the version.

Any suggestion to improve it is welcomed.


mrdeep1:
I don't know how to test this, but otherwise looks fine to me.

PR #1921, opened 2026-03-17
OSS-Fuzz: Revamp block_fuzzer and add corpus seed

arthurscchan:
This PR further simplify the block_fuzzer to make it targets dgram processing directly. It also add 5 seed files with valid CoAP block data and fix the Makefile to zip and copy them for OSS-Fuzz execution.


mrdeep1:
LGTM

PR #1922, opened 2026-03-18
OSS-Fuzz: Add fuzzer targeting coap_async

arthurscchan:
This PR adds a new fuzzer and some seed corpus with valid async package targeting coap_async logic.


tests/oss-fuzz/async_target.c
```diff
@@ -0,0 +1,135 @@
...
+  coap_resource_t *resource = NULL;
+  coap_address_t addr;
+  coap_pdu_t *pdu = NULL;
+
+  if (size < 16)
+    return 0;
+
+  coap_startup();
+  coap_set_log_level(COAP_LOG_EMERG);
+
```
mrdeep1, line 55: To stop packets getting transmitted when running these tests, I would like to see one of both of `coap_debug_set_packet_loss()` and  `coap_debug_set_packet_fail()` used here.  `coap_debug_set_packet_fail() `has recently been added to the develop branch and mimics packet transmission failure.

mrdeep1:
Thanks for doing this.

arthurscchan:
@mrdeep1 Just to confirm my understanding: I have examined the logic in coap_debug_set_packet_loss() and coap_debug_set_packet_fail(). I can see that during packet transmission checking, packet_loss is evaluated first, and packet_fail is only checked if the loss check passes.

Is it correct that with the following configuration:
```
coap_debug_set_packet_loss("50%");
coap_debug_set_packet_fail("100%");
```
I can achieve a 50/50 split where 50% of packets go through the loss path and the other 50% go through the fail path, whilst still guaranteeing that no packets are actually transmitted to the network?

This might provide better fuzzing coverage by exercising both error-handling code paths. WDYT?

mrdeep1:
Your understanding is correct. 

Your solution should give the ~50/50 split and worth going doing.

arthurscchan:
Thanks, I will try making that for all the fuzzers I have created with network packet handling later.

PR #1926, opened 2026-03-19
fix: do not override sequence counter after window processing

BitFis:
During testing it seemed that libcoap is overriding the rcp_ctx->last_seq, which is previously set by `oscore_validate_sender_seq`. My best guess, this should only occure if the window is not initialized, or maybe never?

This leads to strange behavior, that message maybe can be replayed:
1. GET PIV=10 | last_seq = 10
2. GET PIV=7   | (allowed due to PIV window) but sets last_seq = 7
3. GET PIV=10 | is allowed, since last_seq (7) < PIV (10), expected to fail

### DIFF
--- a/src/coap_oscore.c
+++ b/src/coap_oscore.c
@@ -1122,10 +1122,11 @@ coap_oscore_decrypt_pdu(coap_session_t *session,
                                0);
       goto error_no_ack;
     }
-
-    incoming_seq =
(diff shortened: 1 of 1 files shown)


mrdeep1:
Thinks for fixing and reporting this.

I agree with you that this should only be set in the initial state, as it otherwise would have been set by `oscore_validate_sender_seq()`.

PR #1933, opened 2026-03-20
OSS-Fuzz: Fix packet transmission checking for network related fuzzers

arthurscchan:
This PR fixes the packet transmission failing/loss checking configuration for network related fuzzers.


mrdeep1:
LGTM

Issue #1936, opened 2026-03-25
FETCH request with OSCORE option incorrectly rejected with 4.15 when COAP_OSCORE_SUPPORT is disabled


rohit-patil1991:
### Summary

When `COAP_OSCORE_SUPPORT` is disabled at compile-time, a server receiving an
OSCORE-protected FETCH request (e.g., a CoAP Observe subscription from a client
using an external/custom OSCORE stack) is unconditionally rejected with
**4.15 Unsupported Content-Format**.

### Root Cause

In `src/coap_net.c` (function `handle_request`, around line 3861), libcoap
enforces RFC 8132 §2.3.1 by requiring Content-Format to be present in every
FETCH request:

```c
if (pdu->code == COAP_REQUEST_CODE_FETCH) {
    opt = coap_check_option(pdu, COAP_OPTION_CONTENT_FORMAT, &opt_iter);
    if (opt == NULL) {
        /* RFC 8132 2.3.1 */
        resp = 415;
        goto fail_response;
    }
```
## Environment
libcoap develop branch
COAP_OSCORE_SUPPORT disabled
Client using OSCORE-protected CoAP FETCH (Observe subscribe)

## libcoap Configuration Summary
```
knxiot-2.0.11-18-gac9694a43-dirty
```
## Problem Description

When a FETCH request is OSCORE-protected, all inner CoAP options (including
Content-Format) are encrypted inside the OSCORE payload. The outer CoAP message
therefore legitimately carries no Content-Format option. The existing check
has no exception for this case and rejects the request before any application
handler is invoked.

This affects users who:

Disable COAP_OSCORE_SUPPORT (e.g., bring-your-own OSCORE implementation), AND
Receive OSCORE-protected FETCH requests (e.g., Observe subscriptions over OSCORE
as used in KNX IoT / RFC 9176 scenarios).

### Expected Behavior

A FETCH request that carries the OSCORE option (COAP_OPTION_OSCORE, option
number 9) must not be rejected for the absence of Content-Format, because
Content-Format is part of the encrypted inner request and is not visible at the
outer layer.

### Actual Behavior

When `COAP_OSCORE_SUPPORT` is disabled at compile-time, a server receiving an
OSCORE-protected FETCH request (e.g., a CoAP Observe subscription from a client
using an external/custom OSCORE stack) is unconditionally rejected with
**4.15 Unsupported Content-Format**.

### Steps to reproduce

[...]

PR #1938, opened 2026-03-25
FETCH request with OSCORE option incorrectly rejected with 4.15 when COAP_OSCORE_SUPPORT is disabled

rohit-patil1991:
Fixes #[1936](https://github.com/obgm/libcoap/issues/1936)

## Problem

When `COAP_OSCORE_SUPPORT` is disabled and the server receives an
OSCORE-protected FETCH request, `handle_request()` unconditionally
rejects it with 4.15 because Content-Format is absent from the outer
header. Per RFC 8613 §4.1.3.1, Content-Format is an inner option and
MUST NOT appear in the outer OSCORE message.

## Solution

Guard the RFC 8132 §2.3.1 Content-Format check with a prior check for
the OSCORE option. If OSCORE is present the check is skipped; the
encrypted inner request carries the Content-Format and the outer
validation is not applicable.

## Testing

Verified with a client sending an OSCORE-protected FETCH (CoAP Observe
subscription). Before the fix the server returned 4.15; after the fix
the request reaches the application handler correctly.

## Checklist
- [x] Targets `develop` branch
- [x] Single logical change, atomic commit
- [x] `pre-commit` checks pass
- [x] Builds cleanly (CMake)
- [x] References RFC 8132 §2.3.1 and RFC 8613 §4.1.3.1

## Note
If accepted, please consider backporting to the 4.3.x release branch.

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -3859,11 +3859,13 @@ handle_request(coap_context_t *context, coap_session_t *session, coap_pdu_t *pdu
     goto fail_response;
   }
   if (pdu->code == COAP_REQUEST_CODE_FETCH) {
-    opt = coap_check_option(pdu, COAP_OPTION_CONTENT_FORMAT, &opt_iter);
-    if (opt == NULL) {
-      /* RFC 8132 2.3.1 */
-      resp = 415;
-      goto fail_response;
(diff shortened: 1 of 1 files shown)


mrdeep1:
The code change looks fine to me.  I have posted https://github.com/core-wg/corrclar/issues/58 to determine what the experts think should really be happening here.

PR #1940, opened 2026-03-27
oscore: fix OSCORE replay window validation for out-of-order sequence…

ShrutiGunwant17:


mrdeep1:
Thanks for doing the analysis here and coming up with a solution.

I agree that the `- 1` should be removed which then means that `>=` needs to be used.

The overflow guard is a good catch as well.

PR #1942, opened 2026-03-27
OSS-Fuzz: Add Block check fuzzer

arthurscchan:
This PR creates a new fuzzer targets some low level checking of block functions.


mrdeep1:
LGTM

PR #1966, opened 2026-04-21
Features/zephyr 4.4

kr-t:
Hi, I haven't found information about version support, so please let me know if different approach shall be implemented. 

This PR delivers two commits, allowing to use libcoap on zephyr 4.4, which was released last week. Zephyr has a lot of changes, the most notable is they support now MbedTLS 4.x, which already has MbedTLS namespace; additionally, one of the Kconfig configs has changed. I described both changes in their respective commits. 

I have tested these changes both on 4.4 and previous versions, and it seems to not mess up anything. 

Important thing to note regarding new zephyr version:
these are enough to make libcoap support zephyr 4.4 as a library. Due to tons of changes in MbedTLS and PSA crypto, in our local app I had to make tons of config changes to make the application work again as intended. What this means is, that probably for you to support 4.4 and future versions with your sample app (https://github.com/obgm/libcoap/tree/develop/examples/zephyr), you will need to also update those (and decide whether you want to support with the sample app the most recent version, or multiple configs for multiple versions). 

Release notes:
https://docs.zephyrproject.org/latest/releases/release-notes-4.4.html
Migration guide:
https://docs.zephyrproject.org/latest/releases/migration-guide-4.4.html


mrdeep1:
Thanks for doing this.  We certainly want to get support in for 4.4.

These 2 commits should be merged into a single Zephyr commit covering the 2 changes.

I need to think how best to do the MbedTLS configs for 4.3 and 4.4.

kr-t:
Hello, I have squashed the commits into one. Let me know if you'd like anything else. Thanks.

mrdeep1:
LGTM.  

I will separately work on the MbetTLS configs.

PR #1970, opened 2026-04-24
Update version for security scanning

magdalenaszumny:
Update version to latest tag for security scanning


mrdeep1:
Thanks for doing this.

PR #1978, opened 2026-05-05
Draft: Use Zephyr allocators for Zephyr targets

magdalenaszumny:
For Zephyr targets use Zephyr's allocators. Adapted uthash and mbedtls.

### DIFF
--- a/src/coap_mbedtls.c
+++ b/src/coap_mbedtls.c
@@ -119,26 +119,62 @@ typedef mbedtls_md_type_t coap_crypto_md_type_t;
 #define COAP_CRYPTO_MD_NONE     MBEDTLS_MD_NONE
 #endif /* !COAP_USE_PSA_CRYPTO */
 
-#ifdef _WIN32
+/*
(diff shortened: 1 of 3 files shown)


src/coap_mbedtls.c
```diff
@@ -119,26 +119,63 @@ typedef mbedtls_md_type_t coap_crypto_md_type_t;
...
 #include <string.h>
+
+#ifndef mbedtls_malloc
+#define mbedtls_malloc(a) malloc(a)
+#endif
+#ifndef mbedtls_realloc
+#define mbedtls_realloc(a, b) realloc((a), (b))
+#endif
+#ifndef mbedtls_free
+#undef mbedtls_free
```
mrdeep1: The `#undef mbedtls_free `is not needed (was added in because Zephyr was messing things up a few years back).

mrdeep1 (commented):
Apart from one minor change, LGTM.

src/coap_mbedtls.c
magdalenaszumny: Sure, removed!

PR #1979, opened 2026-05-06
OSS-Fuzz: Add a helper for fuzzing

arthurscchan:
This PR adds a new helper header for the OSS-Fuzz fuzzers. The primary purpose is to allow fuzzers to dispatch a correctly formatted CoAP PDU, enabling coverage of code paths that are only reachable after a valid PDU has been received and routed through the dispatch layer.


mrdeep1:
I would prefer that the inline function was moved out of the header file.  See 
[patch.txt](https://github.com/user-attachments/files/27486811/patch.txt) for a suggested way forward (not properly tested).

arthurscchan:
OK. I will followup on that. Thanks for your suggestion.

mrdeep1:
LGTM

PR #1980, opened 2026-05-06
OSS-Fuzz: Fix observe fuzzer

arthurscchan:
This PR improves the observe fuzzer by registering an event callback, adding dispatch via the new coap_fuzz_helper introduced in #1979 to guarantee the registered resource is always reached, and explicitly dispatching GET+Observe PDUs with subscribe and deregister values.

**This PR can only be merged in after #1979 is merged in.**


mrdeep1 (approved):
LGTM

PR #1982, opened 2026-05-06
OSS-Fuzz: Fix block fuzzer

arthurscchan:
This PR improves the block fuzzer by registering an event callback and adding dispatch via the new coap_fuzz_helper introduced in #1979 to guarantee the registered resource is always reached.

**This PR can only be merged in after #1979 is merged in.**


mrdeep1:
LGTM

PR #1983, opened 2026-05-06
OSS-Fuzz: Fix block_check_fuzzer

arthurscchan:
This PR improves the block_check fuzzer by registering an event callback and adding dispatch with options.


mrdeep1:
LGTM

PR #2011, opened 2026-05-08
OSS-Fuzz: Further fix block fuzzer on wellknown handling path

arthurscchan:
This PR fixes the block fuzzer on handling of the well-known/core path.


mrdeep1:
LGTM

PR #2024, opened 2026-05-19
coap_config.h.riot: add CONFIG_LIBCOAP_MAX_STRINGS

fabian18:
When trying to use OSCORE with the current integration into [RIOT](https://github.com/RIOT-OS/RIOT/pull/19889), I had to increase `CONFIG_LIBCOAP_MAX_STRINGS`, but it was missing in `coap_config.h.riot`.

Setting the value in `app.config` did not take effect.


mrdeep1:
Thanks for reporting and fixing this.  I will get the RIOT pr updated shortly to reflect this.

PR #2028, opened 2026-05-20
Add ws_frame fuzzer to be consumed by OSS-Fuzz

DavidKorczynski:
The goal is to increase the code coverage of libcoap while being fuzzing on OSS-Fuzz.


mrdeep1:
LGTM

PR #2032, opened 2026-05-25
Adds extended uri fuzzer to be consumed by OSS-Fuzz

DavidKorczynski:
Extends on the code coverage of the coap_uri.c module, which currently has the code coverage:
https://storage.googleapis.com/oss-fuzz-coverage/libcoap/reports/20260512/linux/src/libcoap/src/coap_uri.c.html


mrdeep1:
LGTM

PR #2033, opened 2026-05-25
Adds extended coap_cache fuzzer to be consumed by OSS-Fuzz

DavidKorczynski:
The goal is to extend code coverage of libcoap. Can confirm from local runs this provides a significant boost to coap_cache.c coverage.


mrdeep1:
Thanks for doing all this coverage work.  LGTM

PR #2042, opened 2026-06-01
Fix data race in coap_resolve_address_info()

michieldwitte:
Steps to reproduce:

```sh
./autogen.sh

CFLAGS="-fsanitize=thread -g -O0" LDFLAGS="-fsanitize=thread" \
  ./configure --enable-thread-safe \
              --disable-doxygen --disable-manpages \
              --disable-dtls --disable-examples

make -j"$(nproc)"
```

`race.c`:
```c
#include <coap3/coap.h>
#include <pthread.h>
#include <stdint.h>
#include <string.h>

static void *worker(void *arg) {
    const char *host = (const char *)arg;
    for (int i = 0; i < 5000; ++i) {
        coap_str_const_t h = {strlen(host), (const uint8_t *)host};
        coap_addr_info_t *info = coap_resolve_address_info(
            &h, 5683, 5683, 0, 0, 0, 1 << COAP_URI_SCHEME_COAP, COAP_RESOLVE_TYPE_REMOTE);
        if (info) coap_free_address_info(info);
    }
    return NULL;
}
int main(void) {
    coap_startup();
    pthread_t a, b;
    pthread_create(&a, NULL, worker, (void *)"127.0.0.1");
    pthread_create(&b, NULL, worker, (void *)"127.0.0.2");
    pthread_join(a, NULL); pthread_join(b, NULL);
    coap_cleanup();
    return 0;
}
```

```sh
cc -fsanitize=thread -g race.c \
   -Iinclude .libs/libcoap-3-notls.a -lpthread -o race
./race
```

This should result in a ThreadSanitizer: data race.

### DIFF
--- a/include/coap3/coap_address.h
+++ b/include/coap3/coap_address.h
@@ -256,6 +256,9 @@ typedef enum coap_resolve_type_t {
  * Resolve the specified @p address into a set of coap_address_t that can
  * be used to bind() (local) or connect() (remote) to.
  *
+ * Note: coap_startup() must have been called first, otherwise @c NULL
+ *       is returned.
+ *
  * @param address The Address to resolve.
  * @param port    The unsecured protocol port to use.
  * @param secure_port The secured protocol port to use.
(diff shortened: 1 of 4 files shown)


mrdeep1:
@michieldwitte Thanks for raising this issue and providing a fix.

However, this may cause an issue with hosts that have constrained stacks.  I would prefer that constrained stacks are recognized with this alternative patch  
[patch1.txt](https://github.com/user-attachments/files/28469927/patch1.txt).

mrdeep1:
Thanks for doing this, and catching the recursive lock within the proxy logic with my initial revision.  LGTM.

mrdeep1:
As an FYI, it it recommended that you create a new branch to create new PRs (using `git checkout -b my-new-patch-name`), rather than just doing the PR update on the develop branch.

PR #2044, opened 2026-06-04
feat: add support for LwIP on microcontrollers

midnight-wonderer:
### Description
When integrating `libcoap` with LwIP (`WITH_LWIP`) on embedded microcontrollers or minimalist SDKs (e.g., bare-metal stacks, lightweight RTOS distributions), compiling libcoap often fails because the target environment's LwIP configuration:
1. Does not declare custom `MEMP_COAP_NODE`/`MEMP_COAP_PDU` pools.
2. Lacks `memp_pools` stats struct members.
3. Lacks support for specific POSIX helpers like `gai_strerror` or certain IPv6 macros.

This PR introduces optional, generic compile-time configuration gates that allow developers to opt-in to standard library memory allocation or compile successfully on lightweight LwIP distributions.

### Changes
1. **Memory Allocator Escape Hatch**:
   Introduce a macro `COAP_LWIP_USE_STDLIB_ALLOC`. If defined, libcoap will use standard `malloc`, `realloc`, and `free` instead of attempting to allocate from LwIP `memp` pools.
2. **Guard Custom Pool References**:
   Add `#ifdef MEMP_COAP_NODE` and `#ifdef MEMP_COAP_PDU` guards around stats and allocations to fall back safely if the pools do not exist in the target LwIP configuration.
3. **POSIX Fallbacks**:
   - Wrap `IN6_IS_ADDR_V4MAPPED` in a standard `#ifndef` check.
   - Define a fallback dummy inline function for `gai_strerror` if the target build indicates it is missing (e.g., via `HAVE_GAI_STRERROR` flag).

### DIFF
--- a/examples/lwip/config/coap_config.h
+++ b/examples/lwip/config/coap_config.h
@@ -124,7 +124,8 @@
 
 #ifndef HAVE_GAI_STRERROR
 /* Minimal fallback for platforms without gai_strerror support */
-static inline const char *gai_strerror(int err) {
+static inline const char *
+gai_strerror(int err) {
   return "getaddrinfo error";
 }
 #endif
--- a/examples/lwip/config/coap_config.h
+++ b/examples/lwip/config/coap_config.h
@@ -122,4 +122,18 @@
 #define HAVE_SNPRINTF 1
 
 
+#ifndef HAVE_GAI_STRERROR
+/* Minimal fallback for platforms without gai_strerror support */
+static inline const char *gai_strerror(int err) {
+  return "getaddrinfo error";
+}
+#endif
+
+#ifndef IN6_IS_ADDR_V4MAPPED
+#define IN6_IS_ADDR_V4MAPPED(a) \
+  ((((const uint32_t *)(a))[0] == 0) && \
+   (((const uint32_t *)(a))[1] == 0) && \
+   (((const uint32_t *)(a))[2] == htonl(0x0000ffff)))
+#endif
+
 #endif /* COAP_CONFIG_H_ */
--- a/examples/lwip/config/coap_config.h
+++ b/examples/lwip/config/coap_config.h
@@ -126,6 +126,7 @@
 /* Minimal fallback for platforms without gai_strerror support */
 static inline const char *
 gai_strerror(int err) {
+  (void)err;
   return "getaddrinfo error";
 }
 #endif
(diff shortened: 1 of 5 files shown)


mrdeep1:
I may be missing something here, but by updating MEMP_USE_CUSTOM_POOLS to 0 in your `lwipopts.h` as per
````
diff --git a/examples/lwip/config/lwipopts.h b/examples/lwip/config/lwipopts.h
index c0646d2c..01b0a9ba 100644
--- a/examples/lwip/config/lwipopts.h
+++ b/examples/lwip/config/lwipopts.h
@@ -68,7 +68,7 @@
 #define MEMP_NUM_SYS_TIMEOUT  10

 /* Set to 0 if custom pools not required (and maybe set MEM_LIBC_MALLOC instead */
-#define MEMP_USE_CUSTOM_POOLS  1
+#define MEMP_USE_CUSTOM_POOLS  0
 #define MEM_SIZE (4 * 1024)

 #if MEMP_USE_CUSTOM_POOLS
````
causes the regular malloc to be used, as in later on in the examples `lwipopts.h`
````
#if ! MEMP_USE_CUSTOM_POOLS
/* Set if you want to use the standard libc for malloc */
#define MEM_LIBC_MALLOC        1
#endif

#if MEM_LIBC_MALLOC
#define HAVE_MALLOC            1
#endif
````

`gai_strerror()` maybe a different story in your environment.

**Comment by midnight-wonderer** [comment]
@mrdeep1 Thanks a lot.
So that's how it should be used. I didn't find it because I'm using the latest release, `v4.3.5b`, rather than `HEAD`.

In my environment, I also had to apply the following patch:

```patch
-#define coap_lock_invert(c,alt_lock,failed) func
+#define coap_lock_invert(c,alt_lock,failed) alt_lock
```

I can see that a lot has improved since the latest tagged release.

I'm closing this issue, but I have one question: which tag or version would you recommend using if we were starting a new project today?

mrdeep1:
> In my environment, I also had to apply the following patch

Yes, that has been fixed in the `develop `branch for some time.

There have been a number of updates for the LwIP support since branch `release-4.3.5`, and only the critical issues have made it into branch `release-4.3.5-patches` which tag `v.3.5b` is based on.  Your needed patch had not made `v4.3.5b`, but has just been added to `release-4.3.5-patches` should a `v4.3.5c` need to be released.

>  which tag or version would you recommend using if we were starting a new project today?

The `develop `branch is pretty stable, I am not expecting much to be changed in the LwIP logic, but there are some other features I want to get into `develop `before the next formal release.

midnight-wonderer:
Awesome, thanks.

PR #2061, opened 2026-06-19
coap_openssl.c: unify tls_client_hello_call_back for PSK and PKI to fix simultaneous operation

mkicherer:
Hello,

if `coap_context_set_psk2` is called after `coap_context_set_pki` to enable both, PKI and PSK, `SSL_CTX_set_client_hello_cb` is called twice and it overwrites the previously registerd callback (`tls_client_hello_call_back -> psk_tls_client_hello_call_back`). Afterwards, PKI-based authentication does not work anymore.

This commit moves the relevant code from `psk_tls_client_hello_call_back` to the `psk_requested` branch in `tls_client_hello_call_back` and calls the latter also in the PSK case. I removed the `#ifdef`-ed code for OpenSSL <1.1.1 as the function seems to be only called for >=1.1.1 anyways.

Thank you!


mrdeep1:
LGTM

PR #2064, opened 2026-06-22
examples: server: Fix const qualifier discard

dimich-dmb:
Compiling with gcc-16.1.1 and glibc-2.43 yields warnings: "initialization discards ‘const’ qualifier from pointer target type".

Indeed, pointer passed to strchr() is `const char *` and memory pointed by return value of strchr() is never modified, so variable to assign return value should be `const char *`.


mrdeep1:
LGTM

PR #2090, opened 2026-07-14
add coap_endpoint_join_mcast_group_intf

mkicherer:
Hello,

I have an application that meticulously tracks interface states and IPs (v4&v6) and creates/deletes endpoints for a specific selection dynamically. As we also use multicast communication, the existing `coap_join_mcast_group_intf` often causes issues like: 

```
setsockopt: Protocol not available
setsockopt: Address already in use
```
as it operates on all registered endpoints of a context.

This commit modifies the existing `coap_join_mcast_group_intf` to enable a new function `coap_endpoint_join_mcast_group_intf` that only sets up multicast for a specific endpoint.

Thank you!

### DIFF
--- a/src/coap_net.c
+++ b/src/coap_net.c
@@ -5518,13 +5518,15 @@ coap_join_mcast_group_intf(coap_context_t *ctx, const char *group_name,
   int ret;
 
   coap_lock_lock(return -1);
-  ret = coap_join_mcast_group_intf_lkd(ctx, group_name, ifname);
+  ret = coap_join_mcast_group_intf_lkd(ctx, NULL, group_name, ifname);
   coap_lock_unlock();
   return ret;
 }
 
 int
-coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
+coap_join_mcast_group_intf_lkd(coap_context_t *ctx,
+                               coap_endpoint_t *single_endpoint,
+                               const char *group_name,
                                const char *ifname) {
 #if COAP_IPV4_SUPPORT
   struct ip_mreq mreq4;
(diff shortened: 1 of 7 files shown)


mkicherer:
I removed the `assert(ctx->endpoint);` before
```
if (!ctx->endpoint)
      return -1;
```
which was also in the original code. I guess an abort with `assert` is not necessary here.

src/coap_net.c
```diff
@@ -5764,6 +5785,18 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
...
 }
 
+COAP_API int
+coap_endpoint_join_mcast_group_intf(coap_endpoint_t *endpoint,
+                                    const char *group_name,
+                                    const char *ifname) {
+  int ret;
+
+  coap_lock_lock(return -1);
+  ret = coap_join_mcast_group_intf_lkd(NULL, endpoint, group_name, ifname);
```
mrdeep1: I suggest you provide the context at this point.
````
 
 if (!endpoint || !endpoint->context)
    return -1;

  coap_lock_lock(return -1);
  ret = coap_join_mcast_group_intf_lkd(endpoint->context, endpoint, group_name, ifname);  
````

File: src/coap_net.c [hunk 2]
```diff
@@ -5535,12 +5537,29 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
...
   coap_endpoint_t *endpoint;
+#if !defined(ESPIDF_VERSION) && COAP_IPV6_SUPPORT && !defined(HAVE_IF_NAMETOINDEX) && !defined(__QNXNTO__)
+  coap_endpoint_t *lookup_endpoint;
+#endif /* !ESPIDF_VERSION && COAP_IPV6_SUPPORT && !HAVE_IF_NAMETOINDEX && !__QNXNTO__ */
   int mgroup_setup = 0;
 
-  /* Need to have at least one endpoint! */
-  assert(ctx->endpoint);
-  if (!ctx->endpoint)
-    return -1;
```
mrdeep1, line 5543: I suggest this is re-instated as ctx is now defined in `coap_endpoint_join_mcast_group_intf()`.

src/coap_net.c
```diff
@@ -5535,12 +5537,29 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
...
+  if (single_endpoint) {
+    if (single_endpoint->proto != COAP_PROTO_UDP)
+      return -1;
+#if !defined(ESPIDF_VERSION) && COAP_IPV6_SUPPORT && !defined(HAVE_IF_NAMETOINDEX) && !defined(__QNXNTO__)
+    lookup_endpoint = single_endpoint;
+#endif /* !ESPIDF_VERSION && COAP_IPV6_SUPPORT && !HAVE_IF_NAMETOINDEX && !__QNXNTO__ */
+    if (!ctx)
+      ctx = single_endpoint->context;
```
mrdeep1: No longer needed as ctx is now defined in `coap_endpoint_join_mcast_group_intf()`.

src/coap_net.c
```diff
@@ -5535,12 +5537,29 @@ coap_join_mcast_group_intf_lkd(coap_context_t *ctx, const char *group_name,
...
+  } else {
+    if (!ctx)
+      return -1;
+    /* Need to have at least one endpoint! */
+    if (!ctx->endpoint)
+      return -1;
```
mrdeep1: Caught by above change revert.

mrdeep1 (changes requested):
Thanks for doing this work.

A lot can be simplified if`coap_endpoint_join_mcast_group_intf()` also provides the context on the call to `coap_join_mcast_group_intf_lkd()`.

src/coap_net.c
mkicherer: Isn't the `!ctx->endpoint` check still necessary?

src/coap_net.c
mrdeep1: Still need this re-instated, as a caller using` coap_join_mcast_group_intf() ` may provide bad information.

src/coap_net.c
mrdeep1: Done by the replaced lines 5540 - 5543.

src/coap_net.c
mkicherer: Ok. I also added the assert again.

mrdeep1:
LGTM

PR #2103, opened 2026-08-01
coap_net.c: Replace %d with PRIu32

kkrentz:
This corrects a format specifier in `coap_net.c`.


mrdeep1:
Thanks for catching this.

PR #2104, opened 2026-08-01
coap_net.c: Avoid divsion by zero if COAP_TICKS_PER_SECOND < 1000

kkrentz:
This fixes a  division by zero on platforms with low timer resolutions.


mrdeep1:
Thanks for fixing this.

PR #2109, opened 2026-08-08
oscore_cbor: fix infinite loop on large CBOR array/map counts

freedom-winds:
CBOR array and map counts can exceed UINT16_MAX. The uint16_t loop counters then wrap and never reach the advertised count, causing a CWE-835 denial of service.

Use size_t loop counters and reject counts that cannot fit in the remaining buffer at the minimum encoding size of one byte per array element or two bytes per map pair. Propagate malformed-value errors and cover large malformed containers and valid values over 64 KiB with regression tests.


mrdeep1:
Thanks for catching and fixing this.

PR #2110, opened 2026-08-10
Stalling Q-Block transfers under packet loss

mkicherer:
Hello,

we are using Q-Block transfers to send files with up to 100 MB to our device. The device is sometimes connected via Wifi. We noticed that if the Wifi signal is weak, the transfers get very slow or even stall completely although other (CoAP) communication is still possible.

We reproduced this behaviour on a Linux system using artificial packet loss. We used the "tc" tool to configure the system to drop packets with an (arbitrary) 10% probability. So almost every batch of blocks is missing one block and recovery is required frequently.

With the help of AI, I was able to fix the hanging and even improve the transfer speed a bit again during packet loss. I must admit that I lack the knowledge about libcoap internals to truly assess the AI patches here. However, this is the result of several hours of testing and trying different approaches so even with AI coming to these results was not easy. Hence, if something looks odd, let me know and I try to come up with something better.

One commit introduces a bitmap that can be used to track all received blocks. This bitmap can be at most 128 KB large but I do not know if this is acceptable for the embedded systems supported by libcoap. Hence, the commit introduces a macro `COAP_Q_BLOCK1_BITMAP_MAX_BYTES` that can be set to limit the acceptable size or even set it to zero. If this is acceptable, let me know if this should be enabled/disabled by default and I wire it into configure and CMake.

I tried to split the changes into multiple commits to ease the review but I can merge them later if you want. I did not test each commit individually.

I asked AI to describe the commits:

52055ff8 Encode large missing Q-Block numbers
----------------------------------------------

Large missing block numbers were encoded as a truncated CBOR uint32, causing the sender to decode and retransmit the wrong block.

* **Encode all uint32 bytes.** Emit the CBOR uint32 marker plus all four big-endian bytes for block numbers at and above 65536.

513f31e4 Keep Q-Block state during recovery
--------------------------------------------

The sender could expire while the receiver was still permitted to send delayed NON 4.08 recovery feedback.

* **Extend the Q-Block state lifetime.** Use the complete NON feedback retry window instead of the ordinary idle timeout.
* **Retain active transfers.** Do not expire Q-Block state while payload work remains active.

12499547 Refresh Q-Block send activity
---------------------------------------

Long Q-Block transfers could appear idle because successful individual payload sends did not update last_sent.

* **Refresh send activity for every payload.** Update last_sent after every successful Q-Block send, preventing an active transfer from expiring.

c76fb64a Handle Q-Block 2.31 Continue responses
------------------------------------------------

The client compared a response code with COAP_RESPONSE_CLASS(231), so valid RFC 9177 2.31 Continue feedback was ignored and the next set waited for a timeout.

* **Compare full response codes.** Use COAP_RESPONSE_CODE(231), allowing valid Continue responses to start the next payload set immediately.

1a851b00 Separate Q-Block recovery transmissions
-------------------------------------------------

Recovery retransmissions previously entered the forward-send queue. They could then reorder or delay new payload sets.

* **Use a dedicated recovery queue.** Keep 4.08-requested blocks separate from forward blocks and restore forward scheduling after recovery flushes.
* **Flush a single recovery block.** A one-block request at a payload-set boundary could remain queued. Treat the recovery path as an explicit send request.

d17a2851 Track Q-Block receives with bitmap
--------------------------------------------

Bounded ranges cannot represent every received block under heavy loss, causing unnecessary retransmissions and preventing exact completion checks.

* **Add optional exact receive tracking.** Use a size-limited bitmap for duplicate, completion, and missing-block checks, with range tracking as a fallback.
* **Release bitmap storage with the transfer.** Free the per-transfer bitmap when receive state is deleted.
* **Detect prior-set gaps immediately.** When later-set traffic arrives, inspect the preceding set and issue 4.08 feedback for real gaps instead of waiting for another timeout.

7e12189f Improve Q-Block missing-block recovery
------------------------------------------------

Range overflow and repeated recovery checks could either lose forward progress or repeatedly send ineffective feedback.

* **Bound missing-block feedback.** Request the earliest real gap and fill remaining feedback capacity with later observed gaps, up to COAP_MAX_PAYLOADS(), without requesting unsent future blocks.
* **Preserve highest-seen receive progress.** Record the largest observed block so recovery can continue after bounded range tracking overflows.
* **Suppress duplicate recovery feedback.** Record outstanding 4.08 requests, clear them only on relevant progress, and retry after the receive timeout.

Thank you!

### DIFF
--- a/include/coap3/coap_block_internal.h
+++ b/include/coap3/coap_block_internal.h
@@ -136,6 +136,7 @@ typedef struct coap_rblock_t {
 #if COAP_Q_BLOCK_SUPPORT
   uint32_t processing_payload_set;
   uint32_t latest_payload_set;
+  uint32_t highest_seen;
 #endif /* COAP_Q_BLOCK_SUPPORT */
   struct coap_lg_range range[COAP_RBLOCK_CNT];
   coap_tick_t last_seen;
--- a/include/coap3/coap_block_internal.h
+++ b/include/coap3/coap_block_internal.h
@@ -117,6 +117,16 @@ struct coap_lg_range {
 #error COAP_RBLOCK_CNT too small
 #endif
 
+#if COAP_Q_BLOCK_SUPPORT
+/*
+ * Maximum bytes used by exact Q-Block1 receive tracking per transfer.
+ * Set to 0 at build time to disable bitmap tracking on constrained devices.
+ */
+#ifndef COAP_Q_BLOCK1_BITMAP_MAX_BYTES
+#define COAP_Q_BLOCK1_BITMAP_MAX_BYTES (128U * 1024U)
+#endif
+#endif /* COAP_Q_BLOCK_SUPPORT */
+
 /**
  * Structure to keep track of received blocks
  */
@@ -255,6 +265,15 @@ struct coap_lg_srcv_t {
   coap_resource_t *resource; /**< associated resource */
   coap_str_const_t *uri_path; /** set to uri_path if unknown resource */
   coap_rblock_t rec_blocks; /** < list of received blocks */
+#if COAP_Q_BLOCK_SUPPORT
+  uint32_t recovery_blocks[COAP_DEFAULT_MAX_PAYLOADS];
+  uint8_t recovery_blocks_used; /**< outstanding Q-Block1 recovery blocks */
+#endif /* COAP_Q_BLOCK_SUPPORT */
+#if COAP_Q_BLOCK_SUPPORT && COAP_Q_BLOCK1_BITMAP_MAX_BYTES > 0
+  uint8_t *q_block1_bitmap; /**< exact Q-Block1 received-block tracking */
+  uint32_t q_block1_bitmap_blocks; /**< number of blocks in bitmap */
+  uint32_t q_block1_bitmap_received; /**< number of received blocks */
+#endif /* COAP_Q_BLOCK_SUPPORT && COAP_Q_BLOCK1_BITMAP_MAX_BYTES > 0 */
   coap_bin_const_t *last_token; /**< last used token */
   uint32_t ref;          /**< Reference count */
 };
--- a/include/coap3/coap_block_internal.h
+++ b/include/coap3/coap_block_internal.h
@@ -189,7 +189,8 @@ struct coap_lg_xmit_t {
   coap_tick_t last_obs; /**< Last time used (Observe tracking) or 0 */
 #if COAP_Q_BLOCK_SUPPORT
   coap_tick_t non_timeout_random_ticks; /** Used for Q-Block */
-  coap_rblock_t send_blocks; /**< list of blocks still to send */
+  coap_rblock_t send_blocks; /**< forward blocks still to send */
+  coap_rblock_t recovery_blocks; /**< Q-Block1 blocks requested by 4.08 */
 #endif /* COAP_Q_BLOCK_SUPPORT */
   uint32_t ref;          /**< Reference count */
 };
--- a/src/coap_block.c
+++ b/src/coap_block.c
@@ -1904,6 +1904,100 @@ check_all_blocks_in(coap_rblock_t *rec_blocks) {
   return 1;
 }
 
+#if COAP_SERVER_SUPPORT && COAP_Q_BLOCK_SUPPORT && \
+    COAP_Q_BLOCK1_BITMAP_MAX_BYTES > 0
+static int
+q_block1_bitmap_has_block(const coap_lg_srcv_t *lg_srcv, uint32_t block_num) {
+  return lg_srcv->q_block1_bitmap &&
+         block_num < lg_srcv->q_block1_bitmap_blocks &&
+         (lg_srcv->q_block1_bitmap[block_num / 8] &
+          (uint8_t)(1U << (block_num % 8)));
+}
+
+static int
+q_block1_bitmap_set_block(coap_lg_srcv_t *lg_srcv, uint32_t block_num) {
+  if (!lg_srcv->q_block1_bitmap ||
+      block_num >= lg_srcv->q_block1_bitmap_blocks)
+    return 0;
+  if (q_block1_bitmap_has_block(lg_srcv, block_num))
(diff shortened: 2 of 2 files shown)


mrdeep1:
Thanks for raising this.  I can see one bug it has fixed, as well as `int `is only 2 bytes on some systems, leading to some potential overflows.

However, I am puzzled by why the use of `lg_srcv->rec_blocks.range` etc. is not doing the same as the proposed bit map.

The use of [coap_debug_set_packet_loss](https://libcoap.net/doc/reference/develop/man_coap_debug_set_packet_loss.html) can be used to set, say, 10% packet loss.

I am assuming from the changes that this is a PUT of a 100Mb file.

I will look at this in detail later.

anyc:
As far as I understood,  the current approach keeps track of COAP_RBLOCK_CNT (=4) ranges with valid blocks. So, if the sender keeps sending blocks after we already have three "holes", these blocks are effectively discarded as we do not know anymore if there was another block missing.

AI explained it this way:

> lg_srcv->rec_blocks.range  is a bounded run-length representation: it records only contiguous received ranges, with  COAP_RBLOCK_CNT  slots (normally four). It works well when loss produces only a few gaps.
> 
> Under sustained/reordered loss, received blocks can form more disjoint ranges than that fixed limit.  blocks_add_entry()  then cannot represent another isolated received block. The range-based recovery path preserves  highest_seen  and asks for earlier gaps, but it cannot retain exact knowledge of every later block that arrived; some already received blocks may be requested again.
> 
> The bitmap is different because it has one bit per expected block:
> 
> • Every received block is recorded exactly, regardless of how fragmented arrival is.
> • Duplicate detection, completion detection, and missing-block feedback are exact.
> • It avoids unnecessary retransmission caused solely by range-table exhaustion

mrdeep1:
I agree that a bit wise solution is a fix.  However, if COAP_RBLOCK_CNT was increased to 5 (and MAX_PAYLOADS stays at 10), then it is possible to track every packet drop within that MAX_PAYLOADS_SET.

[RFC9177 Section 7.2](https://datatracker.ietf.org/doc/html/rfc9177#section-7.2) states
````
Upon receipt of the 4.08 (Request Entity Incomplete) response code, the client SHOULD send the missing
payloads before continuing to send the remainder of the MAX_PAYLOADS_SET and then go into another
NON_TIMEOUT_RANDOM delay prior to sending the next MAX_PAYLOADS_SET.
````
this "go into another NON_TIMEOUT_RANDOM delay" is not happening, and so the server is getting overloaded with missing packets.

I'm trying to track down why this is not happening.

mrdeep1:
I believe that in #2120 this has been correctly addressed for Q-Block1.  Please see if this works for you as well.

Thanks for the ideas from this fix, some of which are in #2120 (possibly slightly modified after testing).

mrdeep1:
@mkicherer @anyc Does #2120 work for you?
A recent push was for something I found with Q-Block2.

mkicherer:
Sorry for the late response, I was on vacation. #2120 also works for me in my test setup. I do not know yet, when I can test this again in the real-world setup but I think this will work there as well. Thank you!

mkicherer:
I just noticed that if I reduce the loss from my usual 10% to 1% or less, the transfer takes much longer. Do you think it would make sense to reduce the timeouts in my case? (Edit: with my kernel-based artificial loss, I still have to test the libcoap-internal loss function)

mrdeep1:
> I just noticed that if I reduce the loss from my usual 10% to 1% or less, the transfer takes much longer.

I need to do the tests using this PR.  With #2120 I get (11M transfer) using default timeouts
````
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data

real    0m0.951s
user    0m0.358s
sys     0m0.226s
q_block_retry [mrdeep1@ubuntu ~/libcoap]
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data -l "10%"

real    56m22.023s
user    0m1.942s
sys     0m1.175s
q_block_retry [mrdeep1@ubuntu ~/libcoap]
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data -l "1%"

real    4m35.503s
user    0m0.823s
sys     0m0.498s
q_block_retry [mrdeep1@ubuntu ~/libcoap]

````
where 1% loss is considerably faster than 10% loss.

mrdeep1:
Using this PR code with this one change
````
diff --git a/src/coap_block.c b/src/coap_block.c
index afc82ebd..7189fd27 100644
--- a/src/coap_block.c
+++ b/src/coap_block.c
@@ -1250,7 +1250,7 @@ coap_add_data_large_internal(coap_session_t *session,
 #if COAP_Q_BLOCK_SUPPORT
     /* Set up send_blocks */
     lg_xmit->send_blocks.used = 1;
-    lg_xmit->send_blocks.range[0].end = (uint32_t)(lg_xmit->data_info->length / chunk);
+    lg_xmit->send_blocks.range[0].end = (uint32_t)((lg_xmit->data_info->length - 1)/ chunk);
 #endif /* COAP_Q_BLOCK_SUPPORT */

     rem = block.chunk_size;

````
I get
````
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data

real    0m0.949s
user    0m0.352s
sys     0m0.210s
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data -l 10%

real    62m26.195s
user    0m2.543s
sys     0m1.266s
$ time examples/coap-client -L7 -m put -f examples/oscore-interop-server coap://127.0.0.1/example_data -l 1%

real    6m44.181s
user    0m0.955s
sys     0m0.561s

````
So not sure why your 1% loss is taking longer than 10% loss.

mkicherer:
Yes, I am sorry, this was an error in my setup script. I hoped that there is something wrong with the slower transfer speed with lower loss probability. I will check if there is something that can be done to improve the speed despite the loss. But I guess this can be closed now. Thank you!

PR #2125, opened 2026-08-24
Restore the request's local address when a triggered async fires

michieldwitte:
Server sessions are keyed on (remote address, local port) only, and session->addr_info.local is overwritten by the destination of every received packet. When a unicast CON request is parked in a coap_async_t for a separate response and a multicast packet from the same peer arrives before the async fires, the re-invoked handle_request() sees a multicast local address, drops the CON per RFC7252 8.1 ("Invalid multicast packet received"), and coap_check_async() frees the async: the response is silently never sent.

Snapshot the local address in coap_register_async() and restore it on the session before the async fires, so the deferred request is handled (and its response source selected) against the address it actually arrived on.

Seen in the field with a CoAP remote control whose GET /endpoint was answered via the async path while it also multicast periodic announcements from the same source port: the device received the empty ACK and then never got the response.

### DIFF
--- a/include/coap3/coap_async_internal.h
+++ b/include/coap3/coap_async_internal.h
@@ -42,6 +42,8 @@ struct coap_async_t {
   void *app_data;                   /**< User definable data pointer */
   coap_app_data_free_callback_t app_cb; /**< callback to call when async is
                                              being released (or NULL) */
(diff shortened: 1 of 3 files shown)


mrdeep1:
Good catch.  Thanks for raising and fixing this.

PR #2146, opened 2026-09-07
Preserve the observe request's local address for notifications

michieldwitte:
Packets from the same peer can overwrite a session's local address with a multicast destination. Save the address when an observation is created or renewed and restore it before generating notifications, so deferred responses snapshot the correct address instead of being dropped as invalid multicast.

Add regression coverage for CON and NON observations, including renewals and the local address captured by a notification's async response.

Validation: 215 unit tests passed (1712 assertions).

### DIFF
--- a/tests/test_session.c
+++ b/tests/test_session.c
@@ -41,6 +41,80 @@ fpeq(const coap_fixed_point_t a, const coap_fixed_point_t b) {
          (a.fractional_part == b.fractional_part);
 }
 
+#if COAP_SERVER_SUPPORT && COAP_ASYNC_SUPPORT
+static coap_address_t observe_expected_local;
+static coap_async_t *observe_async;
+static unsigned int observe_calls;
+
+static void
+observe_address_handler(coap_resource_t *resource COAP_UNUSED,
+                        coap_session_t *s,
+                        const coap_pdu_t *request,
+                        const coap_string_t *query COAP_UNUSED,
+                        coap_pdu_t *response) {
+  observe_calls++;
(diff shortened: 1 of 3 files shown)


tests/test_session.c
```diff
@@ -41,6 +41,80 @@ fpeq(const coap_fixed_point_t a, const coap_fixed_point_t b) {
...
+    coap_pdu_t *request = coap_pdu_init(non ? COAP_MESSAGE_NON : COAP_MESSAGE_CON,
+                                        COAP_REQUEST_CODE_GET, 1, 128);
+    ReturnIf_CU_ASSERT_PTR_NOT_NULL(resource);
+    ReturnIf_CU_ASSERT_PTR_NOT_NULL(request);
+    coap_add_token(request, 1, (const uint8_t *)"t");
+    coap_add_option(request, COAP_OPTION_OBSERVE, 0, NULL);
+    coap_add_option(request, COAP_OPTION_URI_PATH, 12, (const uint8_t *)"address-test");
+    coap_add_resource(ctx, resource);
+    coap_resource_set_get_observable(resource, 1);
+    coap_register_request_handler(resource, COAP_REQUEST_GET, observe_address_handler);
```
mrdeep1, line 79: As a general comment, `coap_add_resource()` should be the last thing called when setting up a resource, as a multi-threaded server could start acting on the partially set up resource on another thread.

mrdeep1 (commented):
Thanks for doing this work.

Only a generic comment to code ordering.

mrdeep1:
LGTM

PR #2149, opened 2026-09-08
Use cryptographic secure entropy for Zephyr

magdalenaszumny:
Use sys_csrand_get which fills the destination buffer with cryptographically secure random data values. sys_rand_get is not considered cryptographically secure.


mrdeep1:
Thanks for fixing this.  LGTM.

PR #2153, opened 2026-09-10
Fix callback data handling for TCP BERT

mkicherer:
Hello,

`assert(rem <= 1024);` is triggered when uploading large data using `coap_add_data_large_request_app` method over TCP. AI suggested this patch which works for me.

It looks like the buffer can be 8 MB large in my case - which might defeat the purpose of the `coap_add_data_large_request_app` approach - but for now it is only important that the application does not abort.

Thanks!


mrdeep1:
Change looks fine to me.

PR #2156, opened 2026-09-15
Mbed TLS: Return the total bytes written over TLS

pptz:
When a message spans multiple mbedTLS records, coap_tls_write() returns only the final mbedtls_ssl_write() result despite sending the entire message. Its caller treats the PDU as partially written and queues an already transmitted suffix.

Update the successful return value to the accumulated byte count. Preserve the existing nonblocking retry and fatal-error handling.

Validated with native mbedTLS 3.6.2 loopback transfers and a focused ASan/UBSan harness covering record boundaries and WANT_READ/WANT_WRITE.


mrdeep1:
Thanks for troubleshooting and fixing this.

PR #2159, opened 2026-09-17
OSCORE: preserve recipient reference on noop attachment

magdalenaszumny:
Prevent oscore_find_context() from decrementing a recipient context reference when the session already owns that context. Add a regression test verifying repeated attachment preserves the reference count.


src/oscore/oscore_context.c
```diff
@@ -388,9 +388,10 @@ oscore_find_context(coap_session_t *session,
         *recipient_ctx = tmp_ctx->recipient_chain;
       }
       tmp_ctx->recipient_chain = *recipient_ctx;
+      const bool already_attached = session->recipient_ctx == *recipient_ctx;
```
mrdeep1: Make this
````
    const coap_bool_t  already_attached = session->recipient_ctx == *recipient_ctx;
````

magdalenaszumny: Thank you! Corrected

mrdeep1:
Thanks for troubleshooting this and coming up with a fix.

PR #2160, opened 2026-09-18
ws: fix NULL dereference on a malformed client-handshake status line

warter666:
In `coap_ws_rd_http_header_client()` the first header (the HTTP status line) is split by `coap_ws_split_rd_header()`, which returns NULL when the buffered line contains no space/tab separator. The second call site checks for NULL, but the first one does not:

```c
value = coap_ws_split_rd_header(session);

if (strcmp((char *)ws->http_hdr, "HTTP/1.1") != 0 ||
    atoi(value) != 101) {
```

If the HTTP header buffer is exactly `"HTTP/1.1"` (no trailing space), `strcmp()` returns 0, the `||` short-circuit does not apply, and `atoi(NULL)` is called - a remote crash in the WS client handshake when a malicious or broken server sends a truncated status line.

Fix: return 0 when `value` is NULL, mirroring the second call site.

The library builds cleanly after the change (cmake -DENABLE_DTLS=OFF, gcc/clang, 0 warnings).


mrdeep1:

PR #2161, opened 2026-09-18
cmake: build/install examples only when the library mode they require is enabled

warter666:
`-DENABLE_CLIENT_MODE=OFF` and `-DENABLE_SERVER_MODE=OFF` are supported CMake options, but the examples are added unconditionally, so both configurations currently fail at link time:

- `ENABLE_SERVER_MODE=OFF`: 186 undefined references (coap-rd.c, coap-server.c, etsi_iot_01.c, tiny.c, oscore-interop-server.c use server-only APIs such as `coap_resource_init`).
- `ENABLE_CLIENT_MODE=OFF`: 8 undefined references (coap-client.c and tiny.c use client-only APIs such as `coap_new_client_session_pki3`).

This adds the missing gates: coap-client requires client mode; coap-rd/coap-server/etsi_iot_01/oscore-interop-server require server mode; tiny requires both. The install(TARGETS) block is gated the same way.

Verified all four combinations (`ENABLE_CLIENT_MODE` x `ENABLE_SERVER_MODE`, DTLS off) now configure, compile, link and install cleanly.


CMakeLists.txt
```diff
@@ -983,32 +983,42 @@ endif()
...
+                     ${CMAKE_CURRENT_LIST_DIR}/examples/oscore-interop-server.c)
+      target_link_libraries(oscore-interop-server
+                            PUBLIC ${PROJECT_NAME}::${COAP_LIBRARY_NAME})
+    endif()
 
-    add_executable(oscore-interop-server
-                   ${CMAKE_CURRENT_LIST_DIR}/examples/oscore-interop-server.c)
-    target_link_libraries(oscore-interop-server
-                          PUBLIC ${PROJECT_NAME}::${COAP_LIBRARY_NAME})
+    if(ENABLE_CLIENT_MODE AND ENABLE_SERVER_MODE)
```
mrdeep1, line 1018: This just needs to be
````
    if(ENABLE_CLIENT_MODE)
````

mrdeep1 (changes requested):
Thanks for doing this work and fixing it.

One minor change required.

warter666:
Done — `tiny` is now gated on `ENABLE_CLIENT_MODE` only, as requested.

mrdeep1:
Thanks for doing this work.  Please squash the 2 commits into a single commit.

warter666:
Squashed into a single commit, as requested.

mrdeep1:
Unfortunately, there still are 2 commits.

warter666:
My previous squash was done incorrectly (it kept a parent inside the PR range), so the branch still showed two commits — sorry about that. The branch is now rewritten properly: a single commit on top of the original base, with the diff unchanged. Verified via the API: the PR now lists exactly one commit.

PR #2162, opened 2026-09-18
block: compile full_match() only when client support is enabled

warter666:
With `-DENABLE_CLIENT_MODE=OFF` (i.e. `COAP_CLIENT_SUPPORT=0`), `full_match()` in `src/coap_block.c` is defined but its only caller is compiled out, producing `-Wunused-function` (fatal in -Werror builds):

```
src/coap_block.c:479:1: error: unused function 'full_match' [-Werror=-Wunused-function]
```

Wrap the static-inline definition in the same `#if COAP_CLIENT_SUPPORT` guard as its caller.


mrdeep1:
Thanks for fixing this. My compiler does not complain about it though....

PR #2163, opened 2026-09-18
debug: fix printf format mismatches reported by -Wextra

warter666:
Three format-string/argument type mismatches in `src/coap_debug.c`, all reported by `-Wall -Wextra` (fatal under -Werror):

1. `msg_code_string()`: `"%u.%02u"` receives `int` expressions (uint16_t promotes to int) - cast both to `unsigned`.
2. `print_content_format()`: `"%d"` receives `unsigned int format_type` - use `%u`.
3. block-size printing: `"%u"` receives the `int` expression `1 << (...)` - make the shift and cast unsigned.

No behavior change on any realistic input; only the int-vs-unsigned variance is fixed.


mrdeep1:
Changes look fine to me.  Not sure what compiler you are using as I do not get this error.

PR #2164, opened 2026-09-18
proxy: guard coap_proxy_get_add_list_entry() against two NULL sessions

warter666:
`coap_proxy_get_add_list_entry(incoming, ongoing, ...)` dereferences `ongoing->context` when `incoming == NULL`. Current in-tree callers always pass exactly one non-NULL session, but the function is a public-ish internal helper and a future caller passing both NULL would dereference NULL immediately. cppcheck flags exactly this (`nullPointer` on `ongoing`).

This adds an explicit early `return NULL` when both arguments are NULL - defensive only, no behavior change for existing callers.


mrdeep1:
This is not part of the Public API, but this defensive code makes sense.

